PR #30, opened 2019-06-28
refactor(platform): add get method in favor of public properties

jgroth:
The old interface was hard to maintain since it violated both the open-closed principle and the
interface segregation principle. Since it for all intents and purposes is a service locator, it
makes more sense to have a clean interface that allows us to add new services without introducing a
breaking change.

BREAKING CHANGE: Removed all public properties, services are now retrieved with the `get` method.
New services can be registered with the `register` method which can also be used by 3rd parties to
register new services on the platform. Core platform service names are exposed in the
`PlatformServiceName` enum.


doc/tutorials/hello-command.mdx
```diff
@@ -77,9 +77,13 @@ lifecycle hook, we can add the code that registers our command and make it useab
 ```typescript
 // src/components/lwc-my-plugin-loader/lwc-my-plugin-loader.tsx
 export class Loader implements LimePluginLoader {
+    private commandBus: CommandBus;
+
     public componentWillLoad() {
-        const helloHandler = new HelloHandler(this.platform.notifications);
-        this.platform.commandBus.register(HelloCommand, helloHandler);
+        const notificationService = this.platform.get(PlatformServiceName.Notification);
```
adrianschmidt: I can't find any example where `PlatformServiceName` is imported. It looks like all the examples of it being used are from loader-components? Will the import be included in the scaffolding created by lime-project? If not, I think we should probably include an example of how to import this.

doc/tutorials/hello-command.mdx
```diff
@@ -77,9 +77,13 @@ lifecycle hook, we can add the code that registers our command and make it useab
 ```typescript
 // src/components/lwc-my-plugin-loader/lwc-my-plugin-loader.tsx
 export class Loader implements LimePluginLoader {
+    private commandBus: CommandBus;
```
adrianschmidt: Hmm… we might need to clarify this example a bit. I was writing a comment about including the import of `CommandBus`, and then I realised we also don't show the `platform` and `context` props, nor the `componentDidUnload`. Aren't those required in the loader-component? I understand that the intention is to only include the relevant code in the example, but I think we're running the risk of confusing the consultants. (Unless I'm the one confused 😂)

adrianschmidt: (I guess it's probably a matter for a different PR though, since this PR isn't really introducing much new material for confusion 🙂)

doc/tutorials/hello-command.mdx
jgroth: I think I have left out most imports in the examples I have written, for good and bad. Too many imports will just become noise when you want to focus on what is important in the example, but it becomes a problem if you don't know where to import stuff from. I don't know what's best, but having all the imports in all examples feels a bit redundant. Maybe a chapter about imports in general?

I also don't think that we have anything from lime-project that generates anything that needs this directly

doc/tutorials/hello-command.mdx
jgroth: Maybe we should have a repo where all the examples can be downloaded fully working? Too much code in the examples will just be clutter, and each snippet cannot be fully functional. It will become hard to see what was actually added in a step

adrianschmidt: Yeah… that might be a good solution.

I guess we'll see. If people don't understand the docs, they'll have to ask us, and then we can improve accordingly.

PR #51, opened 2019-08-07
Notification service report

BregenzerK:
connect to Lundalogik/crm-feature#867


.gitignore
```diff
@@ -4,3 +4,4 @@ dist/
 .docz/
 doc/api/
 .npmrc
+.idea
```
adrianschmidt: ❓🙂

packages/interfaces/src/service/notification-service.interface.ts
```diff
@@ -45,6 +45,18 @@ export interface NotificationService {
...
+
+    /**
+     * Display a non-blocking and non-transient message
+     *
+     * @param {string} message message to display in the banner
+     * @param {ReportOptions} options
+     *
+     * @returns {Promise<boolean>} a promise that will be resolved when the notification is closed.
+     * If the notification has an action button and it was pressed, the promise will resolve to true.
+     * Otherwise, the promise will resolve to false.
```
adrianschmidt: Hmm… good functionality, but do we actually support this in `limel-banner`?

.gitignore
BregenzerK: intellij settings folder.

PR #114, opened 2019-12-30
Tutorial clarifications

### LINKED ISSUE #70: Usages of Event in lime-web-components
When I tried to follow the documentation for "Hello, Event": https://lundalogik.github.io/lime-web-components/versions/latest/#/hello-event#listen-for-command-eventst

I ran into some problems regarding import/missing functions, specifically:  

> getCommandId, CreateLimeobjectDialogCommand and CommandEvent

I have solved the imports using these methods:
`import {getCommandId} from '@limetech/lime-web-components-interfaces/dist/service/commandbus-service';`

`import {CreateLimeobjectDialogCommand} from '@limetech/lime-web-components-commands'`

And adding 'CommandEvent' to the default import from 'lime-web-components'

I'm wondering if this is the correct way to import or if it should be included in the import from 'lime-web-components'? Also it would be great if these imports are specified in the documentation

## Environment

- **lime-web-components** version: 2.1.0
- **lime-project** version: 1.15
- **lime-webserver** version: 12.52.0.117

adrianschmidt:
fix #70


doc/tutorials/hello-event.mdx
```diff
@@ -58,6 +91,10 @@ export class Loader implements LimePluginLoader {
         );
     }
 
+    public componentDidUnload() {
```
BregenzerK: I don't understand how this method help in the tutorial

adrianschmidt: I think it should be in the tutorial, because it's part of the scaffolded file, which also includes a big warning about not removing stuff from it. So I don't think we should be removing stuff from it in our tutorials.

adrianschmidt: I'm not completely sure about `componentDidUnload`, but `componentWillLoad` is required in at least one component because we put some wrapper on it in the web client. That's why we put an empty one in the loader, so there will always be one in one of the loaded components.

I assume the reason for the empty `componentDidUnload` is the same. If it's not needed, I think it should be removed from the scaffold instead.

BregenzerK: I see your point but I also think we should keep them motivated to clean up after themselves. I guess you're referring to this note?

> // NOTE: Do NOT remove this component, it is required to run the plugin correctly.
// However, if your plugin has any code that should run only once when the application
// starts, you are free to use the component lifecycle methods below to do so.
// The component should never render anything, so do NOT implement a render method.

It clearly says not to remove the component. That doesn't mean that you can't keep it clean.

BregenzerK: > I'm not completely sure about `componentDidUnload`, but `componentWillLoad` is required in at least one component because we put some wrapper on it in the web client. That's why we put an empty one in the loader, so there will always be one in one of the loaded components.

We aren't shaffolding any of the lifecycle method in a regular [web-component](https://github.com/Lundalogik/lime-project/blob/master/lime_project/templates/lime-web-component/web-component/component_name.tsx) I thought those 2 methods in the loader are in there as an inspriation for which part of the lifecycle you can affect with the loader.

adrianschmidt: I know we don't scaffold them in a regular component. That's because, at least for `componentWillLoad`, there only has to be one in *one* of the components loaded. I don't know for a fact that that's the case with `componentDidUnload` as well, but until I know, I assume it is. @jgroth might know for sure.

adrianschmidt: I can try removing it and see what happens.

BregenzerK: I just checked the [plugin loader interface](https://github.com/Lundalogik/lime-web-components/blob/master/packages/interfaces/src/lime-plugin-loader.interface.ts). So the wrapper is used by just implementing the interface and from the comment on the `componentDidUnload` I would say you right and it makes no sense to shaffold it.

adrianschmidt: But, it says right there **it must still be present on the component**:

```
    /**
     * This method will never be called since the component will live
     * during the whole lifespan of the application. It must still be
     * present on the component though.
     */
    componentDidUnload(): void | Promise<any>;
```

BregenzerK: I understood that as a reason for why it's part of the interface. Shouldn't it be practically present on the component by implementing the interface? 

> I can try removing it and see what happens.

Excited about any results on that experiment ;) was about to suggest the same

adrianschmidt: > Shouldn't it be practically present on the component by implementing the interface?

No. The interface just specifies methods and properties that must be defined on the object implementing it. An interface cannot define a default implementation, and thus cannot be inherited from.

Removing the `componentDidUnload` method results in the following compiler error:

```
Class 'Loader' incorrectly implements interface
'LimePluginLoader'.Property 'componentDidUnload' is
missing in type 'Loader' but required in type
'LimePluginLoader'.
```

BregenzerK: alright, in that case you're absolutely right and it should also be part of the tutorial!!

adrianschmidt: 🙏

PR #117, opened 2020-01-08
docs(hello world tutorial): fix bug in example by letting `render()` …

adrianschmidt:
…return `<p>` instead of string


Jenkinsfile
```diff
@@ -106,7 +106,7 @@ pipeline {
                             if (DO_RELEASE) {
                                 echo "Found commits that should trigger release! Running release with command ${RELEASE_COMMAND}"
                             } else {
-                                echo "--- FORCING release with command ${RELEASE_COMMAND} ---"
+                                echo "--- FORCING release with command ${RELEASE_COMMAND} BRANCH=\"${env.BRANCH_NAME}\" ---"
```
adrianschmidt: The first build failed, and I added this for debugging purposes. But then the build passed. I don't think it hurts to just keep this here.

PR #121, opened 2020-01-15
docs: improved getting started

jonatanasu-lime:


doc/installation/getting-started.mdx
```diff
@@ -26,7 +28,13 @@ where `<my-component>` is the name of the component.
 
 ## Starting the dev environment
 
-Once a web component has been created, you can start the development by going to the `<plugin_module>/frontend` directory inside the plugin and running:
+Once a web component has been created, you can start the development by going to the `<plugin_module>/frontend` directory inside the plugin and run:
```
adrianschmidt: "running" is actually the correct inflection for "run" here, because it matches "by going" earlier in the sentence. If you prefer, the whole sentence could be changed to:

> Once a web component has been created, go to the `<plugin_module>/frontend` directory inside the plugin and run:

Which is probably better anyway, since the original sentence wasn't great. I suspect "development" was supposed to be "environment" or something…

```suggestion
Once a web component has been created, go to the `<plugin_module>/frontend` directory inside the plugin and run:
```

doc/installation/getting-started.mdx
```diff
@@ -12,7 +12,9 @@ To start developing Lime Web Components, you need to have the following installe
 - node and npm
 - lime-project
 
-We recommend using the latest version of npm and lime-project, and the LTS version of node.
+We recommend using the latest version of npm and lime-project, and the LTS version of Node.js.
+
+`Note: npm is installed with the Node.js installation. Don't check the box "Automatically install the necessary tools..." during installation, since this will install Python 2.7 among other thins you don't need.`
```
adrianschmidt:
```suggestion
**Please note:** npm is installed with the Node.js installation. Do not check the box "Automatically install the necessary tools..." during installation, since this will install Python 2.7 among other things you don't need.
```

PR #140, opened 2020-03-27
fix(decorators): wait for lime properties in componentWillLoad

jgroth:
When a web component is created by Angular it is inserted into the DOM before any props are set.
This will cause the state decorators to fail since the platform prop will be undefined. It will
always work the first time since the components are lazy loaded and Angular will have time to set
the props before the lifecycle hooks run, but if any more components are created after that it will
fail. This change makes sure that the props are set before the lifecycle hooks run.

fix Lundalogik/crm-feature#1004


jgroth:
In order to test this, a new version of lime-crm-components has to be built together with these changes, which then has to be included in a new build with the webclient

packages/decorators/src/index.ts
```diff
@@ -149,6 +153,51 @@ function extendLifecycleMethods(component: Component, properties: Property[]) {
...
+
+/**
+ * Wait for a property to be defined on an object
+ *
+ * @param {LimeWebComponent} target the web component
+ * @param {string} name the name of the property to watch
+ *
+ * @returns {Promise<void>} a promise that will resolve when the property is set on the object
+ */
+function waitForProp(target: LimeWebComponent, name: string): Promise<void> {
```
adrianschmidt: Super-tiny nitpick: I think it would be clearer if the `name` argument was called `property`. Totally up to you if you want to change it or not.

packages/decorators/src/index.ts
```diff
@@ -149,6 +153,51 @@ function extendLifecycleMethods(component: Component, properties: Property[]) {
...
+    const element = getElement(target);
+    return new Promise(resolve => {
+        Object.defineProperty(element, name, {
+            set: (value: any) => {
+                delete element[name];
+                element[name] = value;
+                resolve();
```
adrianschmidt: Ha! Really nice! Nice that it doesn't seem to break Stencil's ability to listen to changes in the value either. It feels like a lot of old-style frameworks like Angular and Knockout would have balked at this 😄

If there's some official resource that led you to this solution, I'd add a link. (And if you came up with it on your own, go buy yourself a beer! 👏👏👏)

jgroth: It was a joint effort between myself and Stack Overflow :smile: 
I found some inspiration for a similar issue, but not the exact same use-case. I knew I wanted to listen for changes on the object, but most solutions suggested to use a proxy object (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy)
We cannot use that here since we are not the one owning the object (Stencil does that). 

I then read about the `defineProperty` method (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty) and did some trail and error to get something working :smile: 

I would have preferred if we could have used the `target` object instead of finding the DOM element, but Stencil has too much magic going on under the hood so it was not possible to use that one :upside_down_face:

adrianschmidt: Ah, yeah, that magic going on under the hood was what I was thinking of when I said it's nice that it doesn't seem to break Stencil's ability to listen for changes 🙂

PR #148, opened 2020-04-06
Testing package

jgroth:


package-lock.json
```diff
@@ -4,7 +4,7 @@
   "dependencies": {
     "@babel/code-frame": {
       "version": "7.0.0",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
+      "resolved": "http://npm.lundalogik.com:4873/@babel%2fcode-frame/-/code-frame-7.0.0.tgz",
```
adrianschmidt, line 7: Your node modules and/or npm cache seem to still be wrong…

jgroth: I tried to clean my cache, hopefully it fixed it :slightly_smiling_face:

adrianschmidt: Seems like it did.

I'm guessing you also removed the package_lock.json file, right? Because it seems like every single package version was updated 😄

That's no big deal though, it's simple to fix.

jgroth: Yeah I had to delete it as well to get it to update

packages/testing/src/spec.ts
```diff
@@ -0,0 +1,164 @@
...
+    addProps(props, element);
+    addListeners(listeners, element);
+    addChildren(element, children);
+
+    element[DESTROY_FUNCTION_NAME] = () => {
+        removeListeners(listeners, element);
+    };
+
+    page.body.appendChild(element);
+    await page.waitForChanges();
```
jgroth, line 60: I will remove this I think. It's probably better if the consumer does the waiting explicitly, that way it will be possible to manipulate the platform services before the first render

packages/testing/package-lock.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "@limetech/lime-web-components-testing",
+  "version": "2.4.0",
+  "lockfileVersion": 1
```
BregenzerK, line 4: since we have a dependency to the interface package, shouldn't that one be in here too? at it added it when locally installed the package.

jgroth: Good catch :+1: 
I'm not quite sure how the lock-files work when using Lerna and a monorepo, but compared to the other sub-packages they seem to have a dependency in the lock-file as well. `npm install` seemed to fix it, I will push a change later today :slightly_smiling_face:

jgroth: Fixed now, I rebased and squashed the commits at the same time
The only change I made was in the lock-file :slightly_smiling_face:

PR #150, opened 2020-04-07
Test fixtures

jgroth:


packages/testing/src/limeobjects/localize.ts
```diff
@@ -0,0 +1,1862 @@
+// tslint:disable:no-duplicate-string
```
jensgustafsson, line 1: I love that we're adding limeoject fixtures but why are we adding `localize` objects as fixtures?

jgroth:
Yeah I can remove the less common fixtures :smile: I just added everything I had in the core database :slightly_smiling_face:

PR #157, opened 2020-04-20
fix(decorators): emit new value when context has changed

jgroth:
The `CurrentLimetype` and `CurrentLimeobject` state decorators are using the context to map the
value of the observable to get the current value och the limetype/object. This works fine when then
component is first created. However, if the context is updated the decorated property will not
receive a new value since the underlying observable will not emit any new values. This change makes
it possible to add a context observable to the limetype and limeobject state services, making them
emit a new value when the context has changed.

BREAKING CHANGE: The `LimePluginLoader` interface now requires the method `componentWillUpdate` to be implemented. This is due to an issue in Stencil where the functionality of the function is not included in the build unless at least one component in the project is using it. Since this fix is using the `componentWillUpdate` function under the hood, it is now required that plugins add it to their projects.

fix Lundalogik/crm-feature#1011


packages/decorators/src/index.ts
```diff
@@ -73,6 +77,11 @@ export function createStateDecorator(
 }
 
 const componentMappings: ComponentMapping[] = [];
+const contexts = new WeakMap<object, LimeWebComponentContext>();
```
jgroth, line 80: :heart: `WeakMap`
We should probably do some refactoring in this file later. We can use `WeakMap` in more places to make the code easier to read/understand :slightly_smiling_face:

packages/decorators/src/index.ts
```diff
@@ -129,13 +138,22 @@ interface Subscription {
...
 function extendLifecycleMethods(component: Component, properties: Property[]) {
     const originalComponentWillLoad = component.componentWillLoad;
+    const originalComponentWillUpdate = component.componentWillUpdate;
     const originalComponentDidUnload = component.componentDidUnload;
     const subscriptions: Subscription[] = [];
 
     component.componentWillLoad = async function(...args) {
         await ensureLimeProps(this);
 
+        const observable = new BehaviorSubject(this.context);
```
Befkadu1, line 148: probably good to name it to `subject` instead of observable. it is used for  to emitted value ?

PR #189, opened 2020-05-14
docs(hello crm tutorial): changing incorrect or missing info

limegga:


doc/tutorials/hello-crm.mdx
```diff
@@ -16,14 +16,14 @@ Create a new component with `lime-project`
 lime-project generate web-component hello-crm
 ```
 
-Just as with our first component, we can remove most of the code in `` and add more as needed. To get started, our component should look like this.
+Just as with our first component, we can remove most of the code in `src\components\lwc-limepkg-my-plugin-hello-crm\lwc-limepkg-my-plugin-hello-crm.tsx` and add more as needed. To get started, our component should look like this.
```
adrianschmidt, line 19: I think we have cleaned up the component template, so I'm not sure this paragraph should even be here anymore? I'll take a look, but I need to wrap up some stuff in our dev-cycle first, since today is the last day of the cycle. I should get to this next week, if not before.

Thanks for the PR, it's very appreciated!

adrianschmidt: Yeah, there's a little bit of stuff left in the scaffold. Let's keep this in 👍

PR #190, opened 2020-05-14
feat(testing): add document limeobject for testing

Befkadu1:
fix Lundalogik/lime-crm-components#27
Test data for document limeobject


packages/testing/src/limeobjects/document.ts
```diff
@@ -0,0 +1,87 @@
...
+        comment: 'Screenshot',
+        type: 40401,
+        company: null,
+        person: null,
+        document: null,
+        coworker: null,
+        deal: null,
+        campaign: null,
+        solutionimprovement: null,
+        _sys_owner: 2901,
```
jensgustafsson, line 20: I think we should remove the object access info keys `_sys_owner`, `sys_group` and `_sys_permissions`.

packages/testing/src/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './spec';
 export * from './platform';
 export { limetypes } from './limetypes';
+export { limeobjects } from './limeobjects';
```
jensgustafsson, line 4: How can this have worked before? Good fix!

Befkadu1: It was not working before.

PR #194, opened 2020-05-26
fix(user-statistics-service): revert version 2.5.0

FredrikWallstrom:


packages/interfaces/src/lime-web-component-platform.interface.ts
```diff
@@ -42,7 +42,6 @@ export enum PlatformServiceName {
     CommandBus = 'commandBus',
     Dialog = 'dialog',
     EventDispatcher = 'eventDispatcher',
-    UserStatistics = 'userStatistics',
```
adrianschmidt, line 45: Since the interface is removed, this is a breaking change, and needs to have a `BREAKING CHANGE:` section in the commit message 🙂

FredrikWallstrom: Even though it is not implemented in lime-webclient?

https://github.com/Lundalogik/lime-webclient/pull/1268

FredrikWallstrom: I updated the commit message :)

PR #203, opened 2020-06-08
feat(interfaces): add aggregates to response

jgroth:
fix Lundalogik/crm-feature#1137


packages/interfaces/src/service/state/limeobjects-service.interface.ts
```diff
@@ -69,4 +69,23 @@ export interface ObjectResponse {
      * Total number of objects that exist matching the query
      */
     totalCount: number;
+
+    /**
+     * Aggregated data
+     */
+    aggregates: AggregationGroups;
```
adrianschmidt, line 76: If `aggregates` is required, isn't this a breaking change?

jgroth: Yes it is! Perhaps it should not have to be required on the interface though. We can make it optional

PR #213, opened 2020-06-23
Update limeobject commnds

Befkadu1:
fix: Lundalogik/crm-feature#1196

This PR is related to the issue https://github.com/Lundalogik/crm-feature/issues/1165


packages/commands/src/limeobject/object-access.command.ts
```diff
@@ -0,0 +1,16 @@
...
+ * Open a dialog for accessing an objects of any type
+ *
+ * The object access dialog can also be used to edit object access for objects of any type.
+ * @id `crm-components.object-access`
+ */
+
+@Command({
+    id: 'crm-components.object-access',
+})
+export class ObjectAccessCommand {
```
jgroth, line 13: I think the name of this could be a little bit more clear. The two already existing commands also opens a dialog and they both are named with dialog in their name. Maybe `OpenObjectAccessDialogCommand`?

packages/commands/src/limeobject/object-access.command.ts
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Open a dialog for accessing an objects of any type
+ *
+ * The object access dialog can also be used to edit object access for objects of any type.
+ * @id `crm-components.object-access`
+ */
+
+@Command({
+    id: 'crm-components.object-access',
```
jgroth, line 11: We should not prefix the id with `crm-components`. The two already existing commands are only prefixed with `limeobject` and that should probably fit quite well here too

packages/commands/src/limeobject/object-access.command.ts
```diff
@@ -0,0 +1,16 @@
+import { Command } from '@limetech/lime-web-components-interfaces';
+
+/**
+ * Open a dialog for accessing an objects of any type
```
jgroth, line 4: We need to rephrase this. It's not very clear what we mean by this

packages/commands/src/limeobject/delete-object.command.ts
```diff
@@ -0,0 +1,15 @@
+import { Command } from '@limetech/lime-web-components-interfaces';
+
+/**
+ * Deletes the object
+ *
+ * @id `crm-components.delete-object`
```
jgroth, line 6: The same issue here with the prefix, we should not have `crm-components` since it's not within the context of this repository

packages/commands/src/limeobject/delete-object.command.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+@Command({
+    id: 'crm-components.delete-object',
+})
+export class DeleteObjectCommand {
+    id: number;
+    limetype: string;
```
jgroth, line 14: We should be consistent and send in the real `Limetype` here instead of just a string. The two already existing commands have the complete object

packages/commands/src/limeobject/object-access.command.ts
```diff
@@ -0,0 +1,16 @@
...
+ * The object access dialog can also be used to edit object access for objects of any type.
+ * @id `crm-components.object-access`
+ */
+
+@Command({
+    id: 'crm-components.object-access',
+})
+export class ObjectAccessCommand {
+    id: number;
+    limetype: string;
```
jgroth, line 15: Same thing here. Send in the `Limetype` and not a `string`

packages/commands/src/limeobject/delete-object.command.ts
```diff
@@ -0,0 +1,15 @@
+import { Command } from '@limetech/lime-web-components-interfaces';
+
+/**
+ * Deletes the object
```
jgroth, line 4: Since this will be visible in the public documentation I think we should be able to write a little bit more than just three words about this command :slightly_smiling_face:

packages/commands/src/limeobject/create-dialog.command.ts
```diff
@@ -26,4 +28,24 @@ export class CreateLimeobjectDialogCommand {
...
+
+    /*
+     * Icon of the dialog
+     */
+    public icon?: string;
+
+    /*
+     * IconColor of the dialog
+     */
+    public iconColor?: string;
```
jgroth, line 50: I'm not sure about these :thinking: They seem very specific to the webclient, and it's not very clear from the documentation what a consumer should enter for the two values. We have been talking about getting addons to work in the desktop client as well, how would these work in there?
Even `subtitle` might be questionable from a platform agnostic point of view

packages/commands/src/limeobject/create-dialog.command.ts
```diff
@@ -1,11 +1,13 @@
 import { Command, Limetype } from '@limetech/lime-web-components-interfaces';
 
 /**
- * Open a dialog for creating a new limeobject
+ * Open a dialog for creating a new limeobject or editing the specific limeobject
```
jgroth, line 4:
```suggestion
 * Open a dialog for creating a new limeobject or editing a specific limeobject
```

packages/commands/src/limeobject/create-dialog.command.ts
```diff
@@ -1,11 +1,13 @@
...
  *
  * The create dialog is implemented as a command so a plugin can easily replace the original dialog with a custom one.
  * Check out the "Hello, Event!" tutorial for a detailed description on how to implement your own create dialog.
  *
+ * This dialog also useful to edit the limeobject that has already created before
```
jgroth, line 9:
```suggestion
 * This dialog also useful to edit a limeobject that already exists
```

packages/commands/src/limeobject/object-access.command.ts
Befkadu1: The type of `limetype` In the below interface is string.
`
export interface LimeWebComponentContext {
    limetype: string | null;
    id: number | null;
}
`
In my implementation, I had `const { id, limetype } = this.context;` And the context is a type of `LimeWebComponentContext`. That is the reason I used a string datatype

packages/commands/src/limeobject/delete-object.command.ts
Befkadu1: I moved the command from `crm-component` and forgot to change the id. 🙏

packages/commands/src/limeobject/create-dialog.command.ts
adrianschmidt: It seems to me like we could probably let lime-webclient set these based on the limetype of the object created. Thoughts?

jgroth: Yeah that's what it does by default currently :slightly_smiling_face: 
I guess these would be to allow a consumer to override that, but I'm not sure if that should be needed :thinking:  Do you have a use case for that now in the history view?

adrianschmidt: Not that I know of. @Befkadu1?

Befkadu1: I did it like that now. I removed the subtitle, icon and iconColor

packages/commands/src/limeobject/delete-object.command.ts
jgroth: This still needs to be updated :slightly_smiling_face:

packages/commands/src/limeobject/object-access.command.ts
```diff
@@ -0,0 +1,14 @@
+import { Command, Limetype } from '@limetech/lime-web-components-interfaces';
+
+/**
+ * Open a dialog to view and edit object access information
+ * @id `crm-components.object-access`
```
jgroth, line 5: This also needs an update

PR #222, opened 2020-07-15
Add dependency on rxjs

adrianschmidt:
`rxjs` is required but not included in dependencies, meaning the module will fail if the consumer
doesn't happen to have `rxjs` among its dependencies or sub-dependencies.


adrianschmidt:
I'm not sure if this import might also require a dependency, or if it's fine because it's just importing a type? @jgroth what do you think?

image

PR #234, opened 2020-08-24
Refactor state decorator factory

jgroth:
requires #232


packages/decorators/src/factory.ts
```diff
@@ -84,24 +79,19 @@ const contextObservables = new WeakMap<
  *
  * @returns {ComponentMapping} mappings for the component
```
BregenzerK, line 81: update return type

BregenzerK (changes requested):

BregenzerK (approved):

PR #238, opened 2020-09-15
fix(interfaces): add totalRelations to limeobject TotalResponse

maltja:
This is to be able to update the tab badge for object-relations

re Lundalogik/crm-feature#1349


packages/interfaces/src/service/state/limeobjects-service.interface.ts
```diff
@@ -87,6 +87,11 @@ export interface ObjectResponse {
      */
     totalCount: number;
 
+    /**
+     * Total number of objects that exist (without any filter applied)
+     */
+    totalRelations: number;
```
jgroth, line 93: We should make this optional, otherwise this is a breaking change

jgroth (changes requested):

jgroth (dismissed):

jgroth (approved):

PR #242, opened 2020-10-15
feat(limeobject): add method for loading schema

maltja:
fix Lundalogik/crm-feature#1476


adrianschmidt (dismissed):
Looks good to me 👍

adrianschmidt (approved):
Works great! 👍

jgroth:
This was a breaking change :smile:

adrianschmidt:
Oh crap… yeah, of course. My bad for not catching that! I'll "fix it" 👍

maltja:
@adrianschmidt @jgroth  How do you handle breaking changes? Do you write it somewhere? 😄

jgroth:
Yeah you should write it in the commit message :slightly_smiling_face: It's very easy if you use `git cz` since it will ask you if it is a breaking change and then ask you to write a message about it. In general though it's better to try and not make it breaking, in this case we made the new method optional on the interface. 

It's a bit weird to have optional methods on interfaces but I think it's good in this case when we are still working on it. We might for instance want to add an `id` argument so you can get a schema for a specific limeobject. We should probably make this method required at some point, but we can do that along with other breaking changes and make the life for consultants a bit easier :slightly_smiling_face:

maltja:
ah okey, thanks for the good explanation 😃

adrianschmidt:
To get the `git cz` command, you need to install commitizen globally: `npm i -g commitizen`.

If you're on a Mac, installing node modules globally may sometimes fail (at least it sometimes does for me), because the access rights on the folder where they're installed are set incorrectly (there's some python script that does that I think). If you run into this, don't just use `sudo` to get around it, instead, the correct ownership and access rights should be set. If you need help fixing that, just ping me 🙂

PR #244, opened 2020-10-15
docs: remove `Installing the plugin` section

jonatanasu-lime:
It's outdated with the move to Poetry. I don't think we need to have the install part of a package or solution documented here.


adrianschmidt:
I think a bunch of this documentation has moved to the lime-crm repo, and should be removed from here. Regardless, this is a good start 🙂

adrianschmidt (approved):

PR #245, opened 2020-10-23
fix(lime-plugin-loader.interface): make deprecated `componentDidUnload` method optional

adrianschmidt:


packages/interfaces/src/lime-plugin-loader.interface.ts
```diff
@@ -27,13 +27,12 @@ export interface LimePluginLoader extends LimeWebComponent {
...
      * This method will never be called since the component will live
-     * during the whole lifespan of the application. It must still be
-     * present on the component though.
+     * during the whole lifespan of the application.
      *
      * @deprecated this method has been deprecated and consumers should
      * use `disconnectedCallback` instead
      */
-    componentDidUnload(): void | Promise<any>;
+    componentDidUnload?(): void | Promise<any>;
```
adrianschmidt, line 35: If this is still required for technical reasons, we need to update that technical reason, as this method has been removed in Stencil 2.

jgroth: It should not be required, so this should be safe :slightly_smiling_face: 
Let's keep it around for some time before we remove it and make a breaking change

jgroth (approved):

packages/interfaces/src/lime-plugin-loader.interface.ts
adrianschmidt: Yeah, I agree 👍

When we do, we should probably make `disconnectedCallback` mandatory as well?

jgroth: Yes, and `connectedCallback` :slightly_smiling_face: 
I think there are a few other breaking changes as well that I have saved up :smile:

PR #247, opened 2020-10-23
Improve limeobject decorator

### LINKED ISSUE #240: Improve Limeobjects decorator
## New feature motivation

The `@Limeobjects` decorator has an option for specifying the limetype of the objects to get from the state. If used, you will only get state updates if any of the objects of that specific limetype has changed. This is great but currently it's very hard to use since you often do not know the limetype, i.e. you have to specify it by a string value

```ts
@Limeobjects({
    limetype: 'this has to be hardcoded :('
})
```

You _could_ specify a getter, but you will not have access to the component so it's still hard to return anything

```ts
@Limeobjects({
    get limetype(): function() {
        return 'no access to the component :(';
    }
})
```

## New feature description

The options for the decorator can be improved to optionally take a function instead that will return the name of the limetype

## New feature implementation

```ts
interface LimeobjectsOptions {
    limetype?: string | ((component?: LimeWebComponent) => string);
}
```

Befkadu1:
fix #240


packages/decorators/src/state/limeobjects.ts
```diff
@@ -2,11 +2,12 @@ import {
     ContextAwareStateOptions,
     PlatformServiceName,
     StateOptions,
+    LimeWebComponent,
 } from '@limetech/lime-web-components-interfaces';
 import { createStateDecorator, StateDecoratorConfig } from '../factory';
 
 export interface LimeobjectsOptions extends StateOptions {
-    limetype?: string;
+    limetype?: string | ((component?: LimeWebComponent) => string);
```
adrianschmidt, line 10: `((component?: LimeWebComponent) => string)`

This part means that you can set the value of `limetype` to be a function that optionally takes a `LimeWebComponent` and returns a `string`, right?

I think this might be a breaking change, because if the implementation only accepts a `string`, it doesn't implement this interface correctly.

One way to test would be to locally install this in lime-webclient master, and run a production build (`npm run build`). Presuming this interface is used in lime-webclient, the build should fail if this change is breaking.

adrianschmidt: Just tried it, and I get this 😞

image

```
ERROR in src/app/services-module/state/limeobjects.service.ts:85:27 - error TS2345: Argument of type 'string | ((component?: LimeWebComponent) => string)' is not assignable to parameter of type 'string'.
  Type '(component?: LimeWebComponent) => string' is not assignable to type 'string'.

85             selector.push(options.limetype);
                             ~~~~~~~~~~~~~~~~
```

Befkadu1: Yea haha, I have not pushed the latest one

Befkadu1: As @jensgustafsson said in the meeting, just wait a little bit and make the PRs and you will test all same time

Befkadu1: But if you are eager to test, this are one of the few lines in my the implementation 

Screenshot 2020-10-26 at 10 42 17

adrianschmidt: I think you misunderstand. I'm not complaining about it not working, I'm saying this is a breaking change in `lime-web-components`, which means we need this to create a new major version, not just a minor version. And when we do that, I think @jgroth has some other stuff we'd like to get merged at the same time, so we don't have to create several new major versions 🙂

Befkadu1: > I think you misunderstand. I'm not complaining about it not working, I'm saying this is a breaking change in `lime-web-components`, which means we need this to create a new major version, not just a minor version. And when we do that, I think @jgroth has some other stuff we'd like to get merged at the same time, so we don't have to create several new major versions 🙂

Aha ok. I talked to @jgroth already, I will contact him when I am ready to do other PRs. Thanks

adrianschmidt: "Breaking change" doesn't mean that it's broken, it just means that the client implementing the interface will have to be updated to work with the new version, because *otherwise* it would be broken 🙂

adrianschmidt: Note that the commit message for this commit (feat(decorators): improve Limeobjects decorator) must be updated before this is merged. It needs to include `BREAKING CHANGE: <description on change>`, or semantic-release will release a new minor version instead of a new major version.

Befkadu1: > Note that the commit message for this commit (feat(decorators): improve Limeobjects decorator) must be updated before this is merged. It needs to include `BREAKING CHANGE: <description of change>`, or semantic-release will release a new minor version instead of a new major version.

Now it is not a breaking change anymore.

packages/interfaces/src/service/state-service.interface.ts
```diff
@@ -31,6 +31,11 @@ export interface StateOptions {
      * The functions will be bound to the web component instance
      */
     filter?: Array<(state: any) => boolean>;
+
+    /**
+     * A component to filter any changes to the state by related limetype.
```
jgroth, line 36: This is way too specific :slightly_smiling_face: Something like this would be better I think
```suggestion
     * The component that is subscribing to the state changes
```

packages/interfaces/src/service/state-service.interface.ts
```diff
@@ -31,6 +31,11 @@ export interface StateOptions {
...
+     */
+    component?: any;
```
jgroth, line 38:
```suggestion
    component?: LimeWebComponent;
```

packages/decorators/src/state/limeobjects.ts
```diff
@@ -2,11 +2,13 @@ import {
     ContextAwareStateOptions,
     PlatformServiceName,
     StateOptions,
+    LimeWebComponent,
 } from '@limetech/lime-web-components-interfaces';
 import { createStateDecorator, StateDecoratorConfig } from '../factory';
 
 export interface LimeobjectsOptions extends StateOptions {
     limetype?: string;
+    getLimetype?: (component: LimeWebComponent) => string;
```
jgroth, line 11: We should add some documentation here describing what this method does

packages/interfaces/src/service/state-service.interface.ts
```diff
@@ -31,6 +32,11 @@ export interface StateOptions {
      * The functions will be bound to the web component instance
      */
     filter?: Array<(state: any) => boolean>;
+
+    /**
+     * The component that is subscribing to the state changes
+     */
+    component?: LimeWebComponent;
```
jgroth, line 39: It does not feel 100% right to add this to the interface, I'm not sure it really belongs here :thinking: 
We should investigate if we can implement this feature without adding the component to an interface that handles options

packages/decorators/src/state/limeobjects.ts
adrianschmidt: Nice! 👍

packages/interfaces/src/service/state-service.interface.ts
Befkadu1: component is removed

jensgustafsson (changes requested):

packages/decorators/src/factory.ts
```diff
@@ -31,12 +40,19 @@ interface Component {
 interface Property {
     name: string;
     options: StateOptions;
+    optionFactory?: (
+        options: LimeobjectsOptions,
+        component: LimeWebComponent
+    ) => StateOptions;
```
jgroth, line 46: We can type this as `type OptionFactory = (options: LimeobjectsOptions, component: LimeWebComponent) => StateOptions` to make it easier to reuse

jensgustafsson: What would be the difference? **Just curious** 😽

jgroth: We are hardcoding it now in two places, so it's to avoid duplicated code :slightly_smiling_face:

Befkadu1: > What would be the difference? **Just curious** 😽

Removes duplicates

jensgustafsson (approved):

PR #248, opened 2020-10-28
feat(save-object): add save-object command

maltja:
fix Lundalogik/crm-feature#1491


packages/commands/src/limeobject/save-object.command.ts
```diff
@@ -0,0 +1,20 @@
...
+} from '@limetech/lime-web-components-interfaces';
+
+@Command({
+    id: 'limeobject.save-object',
+})
+export class SaveLimeObjectCommand {
+    /**
+     * The limetype of the limeobject to save
+     */
+    public limetype: Limetype;
```
jgroth, line 14: I don't think we need this since `LimeObject` has a `getLimetype` method :slightly_smiling_face:

jgroth (changes requested):

packages/commands/src/limeobject/save-object.command.ts
```diff
@@ -0,0 +1,20 @@
...
+
+    /**
+     * The limeobject to save
+     */
+    public limeobject: Partial<LimeObject>;
```
jgroth, line 19: Why is this `Partial`? :thinking:

jgroth:
Clicking the "Save" button works, but the button remains green as if something has been changed after saving is completed

jgroth (dismissed):

jgroth (approved):

PR #251, opened 2020-11-10
feat: add getProperty function

anderssonjohan:
Interface changes for https://github.com/Lundalogik/lime-webclient/pull/1560


packages/interfaces/src/limetype.interface.ts
```diff
@@ -7,6 +7,15 @@ export interface Limetype {
...
     };
     properties: Properties;
+
+    /**
+     * A function to get a property by name
+     *
+     * @param {string} propertyName name of property to get
+     * @throws Will throw an error containing the name of the missing property
+     */
+    getProperty(propertyName: string): Property;
```
jgroth, line 17: We should make this optional to avoid a breaking change :slightly_smiling_face:

packages/interfaces/src/limetype.interface.ts
```diff
@@ -7,6 +7,15 @@ export interface Limetype {
         plural: string;
     };
     properties: Properties;
+
+    /**
+     * A function to get a property by name
```
jgroth, line 12: I don't think we need to state that this is a function :smile: 
```suggestion
     * Get a property by name
```

anderssonjohan: @jgroth https://github.com/Lundalogik/lime-web-components/blob/ac4c3e91ddf40cb832a3d2b91c7caab9f724bc44/packages/interfaces/src/limeobject.interface.ts#L27-L37 :)

anderssonjohan: Was just trying to follow conventions!

packages/interfaces/src/limetype.interface.ts
anderssonjohan: @jgroth fixed

packages/interfaces/src/limetype.interface.ts
```diff
@@ -7,6 +7,15 @@ export interface Limetype {
         plural: string;
```
adrianschmidt, line 7: image

The third commit here will create a bugfix entry in the changelog, but the fix is for a feature that will be added in the same version as the bugfix… 😄

The last two commits should be squashed into the first before we merge this.

(`fix: blah blah` is not the same as `fixup! feat: blah blah`)

anderssonjohan: @adrianschmidt yeah, and I committed the suggestion from github without editing the message there as well. All newer commits are now rebased into the first one so it should be all good!

jgroth (approved):

packages/interfaces/src/limetype.interface.ts
```diff
@@ -7,6 +7,15 @@ export interface Limetype {
...
     properties: Properties;
+
+    /**
+     * Get a property by name
+     *
+     * @param {string} propertyName name of property to get
+     * @throws Will throw an error containing the name of the missing property
+     */
+    getProperty?(propertyName: string): Property;
+
```
adrianschmidt, line 18: There seems to be trailing space here. I'll fix! Then we can get this merged! 🙌

adrianschmidt (approved):

PR #256, opened 2020-12-02
Nested properties

jgroth:
fix Lundalogik/crm-feature#1621


packages/interfaces/src/limeobject.interface.ts
```diff
@@ -25,14 +25,29 @@ export interface LimeObject {
     descriptive: string;
 
     /**
-     * A function to get related data
+     * Get related data of the object
      *
      * @param {Property} property the property of the related object
+     * @deprecated use `getValue` instead
      */
     getRelation(property: Property): LimeObject;
```
adrianschmidt, line 33: I'm thinking this should be made optional since it's deprecated… but maybe that would make this a breaking change for the consumer, since they would then have to check if this method exist before using it, or risk an exception… 🤔

adrianschmidt: I guess we could make it optional in the next major version, and then remove it in the next major after that… 🙂

jgroth: Yeah exactly, we cannot change it now but it's a good idea to make it optional before we remove it :slightly_smiling_face:

adrianschmidt: I'm starting to feel like it's this repo that needs a workflow for "gathering" breaking changes while still releasing minor changes, rather than lime-elements 🙂

We really should get on that I think…

jgroth: Haha yeah, there are quite a few things now that has been deprecated and replaced with something else :slightly_smiling_face:

Befkadu1 (approved):

PR #258, opened 2020-12-10
feat: include route option in command

BregenzerK:


packages/commands/src/limeobject/save-object.command.ts
```diff
@@ -12,4 +12,9 @@ export class SaveLimeObjectCommand {
      * The limeobject to save
      */
     public limeobject: LimeObject;
+
+    /**
+     * route to saved limeobject
+     */
+    public route: boolean;
```
jgroth, line 19: How about we make this optional so it's similar to the create dialog command? :slightly_smiling_face: 
https://github.com/Lundalogik/lime-web-components/blob/master/packages/commands/src/limeobject/create-dialog.command.ts#L27-L30

BregenzerK: :zap:

jgroth (approved):

Issue #264, opened 2021-01-20
lime-web-components: Set up some kind of release flow or branches or something that will let us "collect" breaking changes, and release a bunch all at once


adrianschmidt:
We want to avoid making breaking changes to this interface too often, and both @jgroth and I have things we would like to get done "next time we release a new major version", but there's obviously a rather large risk that we won't remember or have time to do those things at that time. So it would be great to be able to do those fixes, and release them all together at a given time.

Maybe all we need is a branch called "next-major" or something… 🤔

jgroth:
Yeah we should probably add a branch with those changes :sweat_smile: It should not be too hard to keep it up to date, this is a quite slow moving repository with small changes. Doesn't semantic release have automatic handling of brances named `dev`/`alpha`/`beta` etc? Then we would get a released package as well making it easier to test the breaking changes in the wild

adrianschmidt:
Yeah, exactly. First step would be to just add a branch, with some branch protections similar to those for `master` (require PR approval etc). A possible second step would be to set up semantic-release to do prereleases.

`alpha` seems like a good choice to start with, I think 👍

adrianschmidt:
Idea: if possible, add a check that disallows merging any commits marked as breaking into `master` from any branch other than `alpha` and `beta`.

PR #269, opened 2021-02-18
Package refactoring

jgroth:
This is a first step in restructuring this package to make it a bit easier to work with. The current separation into the commands/decorators/interfaces subpackages is weird and we should not split things up based on types like that. If we split things up it's better to do it based on functionality, but I think we can have everything in the same package so consumers will only have to import stuff from `@limetech/lime-web-components`.

I think we can do all this without making any breaking changes by just importing the moved code in the old packages and then re-exporting them and making the old exports deprecated. Code relying on the interfaces from the old packages can still import those types, but will be notified that they are now deprecated and should be imported from `@limetech/lime-web-components` instead. We can then also make all previous semi-breaking changes in the new interfaces instead of breaking the old ones (by semi-breaking I mean those optional properties/methods we have been adding to some interfaces)

All this is very much work in progress and we should probably merge this to an `alpha`, `beta` or `dev` branch as discussed in #264 until we have moved everything we want.

This PR only moves some of the core interfaces like `LimeWebComponent` etc. and the interfaces related to `Limetype`. I also renamed `Limetype` and `Property` to `LimeType` and `LimeProperty` to better align with the Python classes from `lime-core`. The `@Limetypes`/`@CurrentLimetype` decorators should probably be moved to the same `limetype` sub-module, but it seemed hard to move them right now before some other code has been moved as well, so I decided to wait with those for this first step.


adrianschmidt:
Hmm… I've created an `alpha` branch, and set up branch protection for it, but when I was going to configure the release workflow for that branch, I remembered that we're not using semantic-release for this repo, because it doesn't work well with lerna and/or mono-repos. lerna does have some option to create prerelease versions, but I'm not sure how that works, and I'm honestly not too interested in learning it right now, since we'll presumably leave lerna as we're moving away from the mono-repo structure.

So, how about we stop using lerna for versioning (in this PR of course, not on master at the moment), and just use semantic-release for the base package, and then we'll simply have to version the sub-packages manually as needed? I think that would make the whole prerelease setup a lot less complicated…

jgroth:
Yeah that sounds like a good solution :+1: 

Once we have merged all we want we probably are not going to update the sub-packages any longer anyway :thinking:

jgroth:
Ok so I finally managed to get the build green :sweat_smile: I have no idea why it kept failing and I could not reproduce the build errors locally :disappointed: 
It feels a bit weird that I had to do relative imports, but it should hopefully not be an issue

adrianschmidt (approved):
When I try to link this into lime-webclient, I can't get it to build correctly, but I'm thinking it's just as likely to be because the linking isn't getting everything right as there actually being something wrong with the PR, so merging it to `alpha` and releasing a pre-release version seems like the reasonable thing to do 🙂

jgroth:
Yeah linking this package is an issue on master as well, I'm not sure why :thinking: Even when you try to link one of the sub packages it seems to fail :disappointed:

packages/lime-web-components/src/limetype/property.ts
```diff
@@ -0,0 +1,91 @@
...
+    return property && propTypes.includes(property.type);
+}
+
+export function isString(property: LimeProperty) {
+    const propTypes: PropertyType[] = ['string', 'text', 'phone', 'link'];
+
+    return property && propTypes.includes(property.type);
+}
+
+export function isFloat(property: LimeProperty) {
```
Befkadu1, line 87: It would have been better to add docs in the functions

jgroth: Yes there are many places where documentation can be improved

jgroth: This was just a refactor though that moved code from one place to another, improving documentation was a bit out of scope :slightly_smiling_face:

PR #270, opened 2021-03-01
feat(interfaces): add Interval interface

jensgustafsson:
fix Lundalogik/crm-feature#1826

### DIFF
--- a/packages/interfaces/src/interval.interface.ts
+++ b/packages/interfaces/src/interval.interface.ts
@@ -1,0 +1,23 @@
+export interface Interval<T = number | string> {
+    /**
+     * Returns the start of the interval
+     */
+    getLeft(): T;
+
+    /**
+     * Returns the end of the interval
+     */
+    getRight(): T;
+
+    /**
+     * Specifies if the start of the interval is closed (if it includes
+     * the start value of the interval or not)
+     */
+    isLeftClosed(): boolean;
+
+    /**
+     * Specifies if the end of the interval is closed (if it includes
+     * the end value of the interval or not)
+     */
+    isRightClosed(): boolean;
+}
--- a/packages/interfaces/src/index.ts
+++ b/packages/interfaces/src/index.ts
@@ -6,5 +6,6 @@ export * from './lime-web-component-platform.interface';
 export * from './lime-web-component.interface';
 export * from './query.interface';
 export * from './limeobject.interface';
+export * from './interval.interface';
 
 export * from './service/index';


packages/interfaces/src/interval.interface.ts
```diff
@@ -0,0 +1,23 @@
+export interface Interval<T = number | string> {
```
adrianschmidt, line 1: If we add a docbloc comment here, will that show up in the docs as the description for the whole interface? Asking because it seems to me like this might need at least _some_ description for people to know what it's for.

```suggestion
/**
 * <description here>
 */
export interface Interval<T = number | string> {
```

packages/interfaces/src/interval.interface.ts
```diff
@@ -0,0 +1,23 @@
...
+    /**
+     * Returns the start of the interval
+     */
+    getLeft(): T;
```
adrianschmidt, line 5: Is this interface modelled on some precedent, or did we come up with the names ourselves?

I assume using "left" and "right" is an attempt at keeping the interface type-agnostic, but I'm wondering if it doesn't confuse things more than it helps… especially since we then explain the meaning of them using "start" and "end"… 🤔

adrianschmidt, line 5: Separate question regarding the `get`-part (and here I'm treading in areas where I may just be wondering because I don't know anything about what the interface will be used for, so if I'm just asking stupid questions, just tell me to go away 😂):

Why `get`? Is it important that the user of the interface can't change the values?

packages/interfaces/src/interval.interface.ts
```diff
@@ -0,0 +1,23 @@
...
+    /**
+     * Returns the end of the interval
+     */
+    getRight(): T;
+
+    /**
+     * Specifies if the start of the interval is closed (if it includes
+     * the start value of the interval or not)
+     */
+    isLeftClosed(): boolean;
```
adrianschmidt, line 16: Similar to my question about using left/right rather than start/end or similar: is there some precedent for using "closed" here? I guess another way to ask the same question is: Are (some) people already familiar with 'closed' in the meaning it's used here? Because I'm not, and I feel like I'd have some trouble figuring it out, if not for the description explaining it. So I wonder if there might be some simpler name, like `hasLeft` (or `hasStart`, `hasBeginning`, or whatever might be suitable if we were to change from left/right to something else)?

jgroth: The left/right and open/closed terminology comes from mathematics. Whether that's a good fit here or not is debatable I guess :sweat_smile: 
https://en.wikipedia.org/wiki/Interval_(mathematics)

adrianschmidt: Ah, ok, but then there's exactly the type of precedence I was wondering about 👍

Since the interface is type-independent, I think the mathematics terminology makes sense to use. (I apparently misunderstood the meaning of of "includes the start value", but got it cleared up when reading about the terms "open" and "closed" in that Wikipedia article 👍)

jensgustafsson: Johan asked me why I wanted to add the `Interval` interface here. I couldn't come up with a good answer and after a discussion we decided that I/we should move this in to `lime-crm-components` instead.

adrianschmidt: Yeah, it's easier to change while it's private, and we can always make it public by moving it here later if we feel that's the right thing to do 👍

jgroth:
Closing this since we moved it to crm-components

PR #274, opened 2021-03-11
feat(filter-service): implement filter service interface

Befkadu1:
fix: Lundalogik/crm-feature#1913


packages/interfaces/src/service/filter-service.interface.ts
```diff
@@ -0,0 +1,11 @@
+import { Filter } from './../query.interface';
+export interface FilterService {
```
jgroth, line 2: This is a state service so we should extend the `StateService` interface. We should probably also move this to the `state` sub-folder to make this more clear

packages/interfaces/src/service/index.ts
```diff
@@ -8,3 +8,4 @@ export * from './commandbus-service';
 export * from './dialog-service.interface';
 export * from './task-service.interface';
 export * from './event-dispatcher-service';
+export * from './filter-service.interface';
```
jgroth, line 11: Move this to the state service instead

packages/interfaces/src/service/filter-service.interface.ts
```diff
@@ -0,0 +1,11 @@
...
+    /**
+     * Saves the new filter to the database
+     *
+     * @param filter a filter to save
+     *
+     * @returns {Promise<void>} a promise that will be resolved when the filter is saved
+     */
+    saveFilter(filter: Filter): Promise<void>;
```
jgroth, line 10: We already know that we have a filter and the service is also named `FilterService`, so I think it should be enough to just name this `save`

jgroth (dismissed):

PR #275, opened 2021-03-12
Platform test errors

jgroth:


packages/testing/src/platform/platform.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        beforeEach(() => {
+            platform.register('foo', new MyService());
+        });
+
+        it('the service works as expected', () => {
+            const service: MyService = platform.get('foo');
+
+            expect(service.getData()).toEqual('foo');
```
adrianschmidt, line 35: Looks like the service _name_  and the service method's _return value_ are both `'foo'`? That makes the code harder to understand. I suggest changing those to separate values, so it's easier to follow the code and know which is which.

packages/testing/src/platform/platform.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+        });
+
+        it('unknown methods returns the arguments', () => {
+            const service: any = platform.get('foo');
+
+            expect(service.foo('bar')).toEqual(['bar']);
```
adrianschmidt, line 41: Another `foo`, although at least not `'foo'` this time 😅

Maybe this can be called `service.unknownMethod` or something instead?

packages/testing/src/platform/platform.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+        beforeEach(() => {
+            const map = new Map();
+            platform.register('map', map);
+        });
+
+        it('works without throwing errors', () => {
+            const service: Map<string, any> = platform.get('map');
+
+            expect(() => service.has('foo')).not.toThrow();
+        });
```
adrianschmidt, line 55: "when registering a Map it works without throwing errors"

It's not the registering of the Map we check for errors though, it's the call to `service.has`. The description should reflect that.

adrianschmidt:
Good job! I think all the things I commented on were pretty minor, but easy to fix with decent return on investment 😊

jgroth:
Thank you for the review, I fixed your comments and squashed it since they were quite minor. The diff should be this one
https://github.com/Lundalogik/lime-web-components/compare/2105f65c30d4c9c9053dcfc99093d2b3f8616779..bc20f4ad6e61dc5706e4b894439df378918c7383

adrianschmidt (approved):

PR #277, opened 2021-03-18
feat(filter-service): implement update to the filter service interface

Befkadu1:
fix Lundalogik/crm-feature#1931

### DIFF
--- a/packages/interfaces/src/service/state/filter-service.interface.ts
+++ b/packages/interfaces/src/service/state/filter-service.interface.ts
@@ -10,4 +10,13 @@ export interface FilterService extends StateService {
      * @returns {Promise<void>} a promise that will be resolved when the filter is saved
      */
     save(filter: Filter): Promise<void>;
+
+    /**
+     * Updates the existing filter
+     *
+     * @param filter a filter to update
+     *
+     * @returns {Promise<void>} a promise that will be resolved when the filter is updated
+     */
+    update(filter: Filter): Promise<void>;
 }


packages/interfaces/src/service/state/filter-service.interface.ts
```diff
@@ -10,4 +10,13 @@ export interface FilterService extends StateService {
...
     save(filter: Filter): Promise<void>;
+
+    /**
+     * Updates the existing filter
+     *
+     * @param filter a filter to update
+     *
+     * @returns {Promise<void>} a promise that will be resolved when the filter is updated
+     */
+    update(filter: Filter): Promise<void>;
```
jgroth, line 21: Why do we need this? :slightly_smiling_face: 
Would not the `save` method cover all the use cases that this method also has? What is the difference between them? When should I use `save` and when should I use `update`?

Befkadu1: The update method uses `PUT` request on the webclient. Or we need to send the action type (`update` or  `create`) to the method if we want to use only one method, `save`

jgroth: Yes, but that is an implementation detail that a consumer of this interface should not have to be aware of. They just want to save the filter and should not have to keep track of what method to use. I think we should be able to implement this feature with just the `save` method

Befkadu1: This PR is closed, not needed anymore

PR #278, opened 2021-03-19
feat(filter-service): implement delete method to the filter service i…

Befkadu1:
…nterface

fix: Lundalogik/crm-feature#1936


packages/interfaces/src/service/state/filter-service.interface.ts
```diff
@@ -10,4 +10,12 @@ export interface FilterService extends StateService {
...
      */
     save(filter: Filter): Promise<void>;
+    /**
+     * Deletes the filter
+     *
+     * @param id a filter id to delete
+     *
+     * @returns {Promise<void>} a promise that will be resolved when the filter is deleted
+     */
+    delete(id: string): Promise<void>;
```
jgroth, line 20: This is a breaking change and needs to be marked as such in the commit message.
However, a better approach right now is to make this method optional and then we can make it mandatory the next time we make a breaking change.

packages/interfaces/src/service/state/filter-service.interface.ts
```diff
@@ -10,4 +10,12 @@ export interface FilterService extends StateService {
...
      */
     save(filter: Filter): Promise<void>;
+    /**
+     * Deletes the filter
+     *
+     * @param id a filter id to delete
+     *
+     * @returns {Promise<void>} a promise that will be resolved when the filter is deleted
+     */
+    delete?(id: string): Promise<void>;
```
jgroth, line 20: What do you think about passing the actual filter instead of just the id? It will make the method more similar to `save`

Befkadu1: Is it only to make it similar to the `save` method or you have other reason? The only thing we need is id in this case eller?

jgroth (approved):

PR #279, opened 2021-03-25
feat(route-service): add optional params when routing to table

hannahu:
fix Lundalogik/crm-feature#1891


packages/interfaces/src/service/route-service.interface.ts
```diff
@@ -23,6 +24,7 @@ export interface RouteService {
      *
      * @param {string} limetype the limetype of the table to route to
      * @param {Expression} filter filter expression to apply to the table
+     * @param {Limetype[]} transferFrom the limetypes the filter has been transferred from
      */
-    table(limetype: string, filter?: Expression);
+    table(limetype: string, filter?: Expression, transferFrom?: Limetype[]);
```
jgroth, line 29: I think it would be a good idea to wrap the last argument in something called `options` or `params` or similar. The reason is that we don't know if we want to add other parameters in the future, and it will be hard to use this if the argument list grows too long. Having a generic param argument would also allow addons to send in stuff if they need it for a custom view etc.

`filter` should probably also be part that `param`/`option` argument, but we cannot change that now since it's a breaking change. We can merge them later if we rename the `table` method to something else, `table` will probably not make much sense once other views are added :sweat_smile: 

Also, adding the full `Limetype` object will be hard to serialize to put in the URL, the name of the limetype is probably enough and then it's up to the component to get the full limetype object

What do you think?

packages/interfaces/src/service/route-service.interface.ts
```diff
@@ -23,6 +23,7 @@ export interface RouteService {
      *
      * @param {string} limetype the limetype of the table to route to
      * @param {Expression} filter filter expression to apply to the table
+     * @param {Object} params params to use when routing to the table
      */
-    table(limetype: string, filter?: Expression);
+    table(limetype: string, filter?: Expression, params?: any);
```
jgroth, line 28: Specifying this as `any` will allow the consumer to send in `string`, `number` etc. which will be impossible to handle. The correct type should be `Record<string, any>`

jgroth (approved):

PR #280, opened 2021-04-14
feat(limetype.interface): create PropertyOption interface and use it …

Befkadu1:
…in Property interface

fix Lundalogik/crm-feature#1828


packages/interfaces/src/limetype.interface.ts
```diff
@@ -39,19 +39,21 @@ export interface Property {
...
     required: boolean;
     type: PropertyType;
-    options?: Array<{
-        key: string;
-        inactive: boolean;
-        text: string;
-        order?: number;
-        id?: number;
-    }>;
+    options?: PropertyOption[];
```
adrianschmidt, line 42: Ah! That was a simple solution! 😊👍

adrianschmidt (dismissed):

packages/interfaces/src/limetype.interface.ts
```diff
@@ -39,19 +39,21 @@ export interface Property {
...
     relation?: {
         getLimetype: () => Limetype;
         getBackreference: () => Property;
     };
 }
 
+export interface PropertyOption {
```
adrianschmidt, line 49: Last minute note from @jgroth: we should check if there's a corresponding property in the backend, and if so, name this the same as in the backend.

adrianschmidt: It's apparently just called `Option` inte the backend, so I think `PropertyOption` is just fine 👍

adrianschmidt (approved):

packages/interfaces/src/limetype.interface.ts
jgroth: Just FYI :slightly_smiling_face: 
`Property` will be renamed to `LimeProperty` in the next major release to better align with the corresponding type in Python (https://github.com/Lundalogik/lime-web-components/blob/alpha/packages/lime-web-components/src/limetype/property.ts#L4)

The corresponding type for this is `Option` but that might be too generic here so `PropertyOption` could work, another alternative is `LimePropertyOption` :thinking:

PR #281, opened 2021-04-14
fix(command-bus): add getHandler type to command bus service

sampsonbryce:


packages/interfaces/src/service/commandbus-service.ts
```diff
@@ -31,6 +31,15 @@ export interface CommandBusService extends CommandHandler {
...
     isSupported(commandId: CommandIdentifier): boolean;
+
+    /**
+     * Get a handler associated with a command
+     *
+     * @param {CommandClass} commandClass The command class
+     *
+     * @returns {CommandHandler} the handler for the command class
+     */
+    getHandler(commandClass: CommandClass): CommandHandler;
```
jgroth, line 42: This is a breaking change and we would like to avoid those if possible. I suggest making this optional for now and then we make it mandatory on the next major release

jgroth (approved):

PR #286, opened 2021-06-10
remove week and datetime to PropertyType interface

Befkadu1:


hannahu (approved):

packages/interfaces/src/limetype.interface.ts
```diff
@@ -67,11 +67,9 @@ export type PropertyType =
     | 'percent'
     | 'time'
     | 'date'
-    | 'datetime'
     | 'year'
     | 'quarter'
     | 'month'
-    | 'week'
```
adrianschmidt, line 74: 1. Weren't these _just_ added?

And

2. This is a breaking change, and should have been marked as such.

Befkadu1: It was added it [here](https://github.com/Lundalogik/lime-web-components/pull/285) a couple of hours ago and removed it, it is like reverting it . They will be implemented when we need them.

adrianschmidt: Well… yeah, it's kind of like reverting, only that you "reverted" something in a public API, which is a breaking change 🙂

https://apifriends.com/api-management/an-api-is-a-promise/

Now, I'm sure no one had the time to update to the last version and start implementing functionality on that new API, so it shouldn't be any major problem this time. But if we want to add things to our public APIs, we should aim to have considered the change well enough that we don't just change our minds a few hours later. And if we _do_ change our minds a few hours later (which might still happen, and that's ok) we need to take responsibility for that, and properly document that we're making a breaking change by removing something from the API.

PR #287, opened 2021-06-14
Added length as optional property on interface Property

TommyLindh2:
The `length` property is the maximum number of characters allowed in a string property i.e. `text` `string` `phone` `link` e.t.c.


adrianschmidt:
@TommyLindh2 Please add at least one sentence about the rationale for the change to the PR description. I assume this value is for getting the maximum allowed length of the property? Or is it for getting the current length?

TommyLindh2:
I've updated the description now

adrianschmidt (approved):

PR #289, opened 2021-09-05
feat: adds url query params decorator

hannahu:
fix Lundalogik/crm-feature#2281


packages/decorators/src/state/query-params.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+export function QueryParams(name?: string) {
+    const config: StateDecoratorConfig = {
+        name: PlatformServiceName.Route,
+    };
+
+    const options: RouterOptions = { target: 'queryParams' };
+
+    if (name) {
+        options.name = name;
```
jgroth, line 26: How does this compile? :sweat_smile: 
`name` is neither part of `StateOptions` or `RouterOptions`

packages/decorators/src/state/query-params.ts
```diff
@@ -0,0 +1,30 @@
+import {
+    PlatformServiceName,
+    StateOptions,
+} from '@limetech/lime-web-components-interfaces';
+import { createStateDecorator, StateDecoratorConfig } from '../factory';
+
+export interface RouterOptions extends StateOptions {
+    target?: string;
```
jgroth, line 8: This should be better documented so that we remember what it's used for.

However, I think we can scrap it and just use the name of the param instead, e.g
```ts
export interface RouterOptions extends StateOptions {
    queryParam?: string | boolean;
}
```

If `queryParam` has been set on the options, we then know that we want to select a query parameter and can decide inside the service how to do that. The decorator then does not need to know any inner details about the service and the shape of the state

hannahu: Yes!

packages/decorators/src/state/query-params.ts
hannahu: I have super powers

packages/decorators/src/state/query-params.ts
hannahu: But do you think it's correct to call it `RouterOptions` and have it here? Maybe in another place?

jgroth: It should probably be moved to the `interface` package since it's part of the service.

However, it's a bit weird because stuff like this should belong together. It has ended up this way because of our poor project structure in this repo where we have grouped things by type rather than functionality. It will be fixed in the next major update (work started in the `alpha` branch)

packages/decorators/src/state/query-params.ts
jgroth: :superhero:

PR #290, opened 2021-09-09
feat: make user-data service public

anderssonjohan:
In various limepkg's we want to persist some state between page views/reloads and why not use what we already got there?

This already works, but since the interface is private, we don't dare to use this service for now.

```javascript
ud = this.platform.get('webclient.user-data')
ud.set('limepkg-awesome', 'stuff')
ud.get('limepkg-awesome') // -> "stuff"
JSON.parse(localStorage['lime-crm.state']).userData['limepkg-awesome'] // -> "stuff"
```

A limepkg/solution loader component could even do its own state migration like we already do in the webclient:
https://github.com/Lundalogik/lime-webclient/blob/a6752ef8dc710929b0b3c4c92b1363ec0ddcd384/frontend/webclient/src/app-ng1/common/reducers/state-migration.ts#L32

Note: With this rather naive "fix" we assume that using this key-value store every package/solution is responsible to do proper prefixing. The webclient already uses `webclient.` for userData and it would be expected that solutions/limepkgs use `solution-xyz.` and `limepkg-xyz.` as their key prefix.


packages/interfaces/src/service/state/user-data-service.interface.ts
```diff
@@ -4,10 +4,6 @@ import { StateService } from '../state-service.interface';
  * Handle fetching and saving user specific data, e.g. settings or other state
  * data that should persist between sessions
  *
- * This is an experimental service and is subject to change in the future. The
- * API should not be considered stable
- *
- * @private
```
adrianschmidt, line 10: @jgroth What do you say?

I have a few thoughts:

1. It might be useful to require a prefix as a separate string from the key. That way, it becomes more obvious that you "bloody well better" supply a prefix 😆 We can write all the documentation about providing prefixes we want to, but if all we require is a single `key: string`, we all know some people are just going to send in stuff like `"state"` anyway…
2. I agree that, with the above change, this interface seems plenty useful as it is, and should be exposed. However, if we feel like we want to be able to make changes to the interface that we use in lime-webclient and lime-crm-components, couldn't we keep this private and just expose another almost identical interface, that add-ons can use?
3. If I recall correctly, while this interface is called `UserDataStateService`, the storage isn't actually limited to the logged in user? If that's the case, I think we should either fix that while we're at it, or we the interface we expose should use a different name.

Last one is a question: Is stuff sent in here was stored in localStorage or in the db on the server? I know I've asked this before, but I can't remember the answer… 🤦‍♂️

jgroth: Copy/paste from a thread we had ongoing on Slack :slightly_smiling_face: 

> I'm not sure the current implementation of the service is ready for wide spread usage yet. At the moment it just puts stuff in localStorage and does not care about which user is doing it. If you log in as another user or switch to another database it will just get what's in localStorage and return that :stuck_out_tongue: We should be able to fix that without the backend stuff, but it would be really nice to have that as well before making it public. Team up for Hack Tuesday maybe? :slightly_smiling_face:

adrianschmidt: Yeah, I think separating the data based on logged in user is a must before we release the interface, if nothing else, then for the simple fact that the interface name is a lie until we fix that. I don't necessarily think there has to be any real _security_ implemented, in the sense that one user shouldn't be able to access what another user stored, I assume they can do that regardless, since it's stored in localStorage, right? Unless we encrypt the data. The documentation should be _very_ clear about the lack of security though…

jgroth: Yeah, but if we store it in the backend instead we can scrap localStorage altogether, no need to store in two places :slightly_smiling_face:

adrianschmidt: Yeah, of course… I didn't mean to say we should keep storing it in localStorage once we've added server-side capability to it… 😄

anderssonjohan: The only reason I see to keep a "buffer" in localStorage is that changing it from what it is today to server-side will change the interface to be async. Usually we want to show some kind of spinner/loading indicator when doing requests to the server and I don't know where we use this interface in the webclient today.. that is, where we may need to add some spinner etc if we change it to be async + server-side.

adrianschmidt: Ah… good point, but not in the way you mean 😝

Uses in lime-webclient, we have control over, so we can find and change those when the implementation is changed. But what you said is a pretty good argument for why we haven't published this api yet, because as soon as we do, going from sync to async is a big fat breaking change…

So, apart from separating data based on logged in user, I think it's also a must to change this interface to be async (that is, to return a promise), before we make it public.

jgroth:
If it's ok I will close this in favor of #294, where we added a decorator that can be used in components to subscribe to the data as well :slightly_smiling_face:

PR #291, opened 2021-09-10
test(tesing/company): add fixtures for company limetype

Befkadu1:
add mothercompany, mothercompanylime and subsidiary properties to company limetype for testing


packages/testing/src/limetypes/company.ts
```diff
@@ -559,6 +559,65 @@ export const company = {
...
+            has_sql: false,
+            options: [],
+            acl: {
+                read: true,
+                update: true,
+                create: true,
+                delete: true,
+            },
+        },
+        mothercompanylime: {
```
jgroth, line 579: I'm not sure why both these are added. This was added to test relations to the same limetype right? What is the purpose of the `string` field? 

My suggestion is to remove the field `mothercompany` and rename the field `mothercompanylime` to `parentcompany` (I think parent company is the correct term here, or holding company)

Befkadu1: In my db I found `mothercompany`  with string type that is why i added.
I have also `mothercompanylime`. May be I have an old db?

jgroth: Aha, yeah I don't know haha :smile: I did not know you took these from an actual database :slightly_smiling_face:

Befkadu1: Anyway I implemented your suggestions

jgroth: I did some googling and it seems parent/holding company is a much more common term than mother company, but I'm neither a linguist or accountant :smile:

jgroth (dismissed):

jgroth (approved):

PR #294, opened 2021-09-22
feat(decorators): add decorator for accessing user data

jgroth:
fix Lundalogik/hack-tuesday#190


packages/decorators/src/state/user-data.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+/**
+ * Gets an object with all user data where key is used as key.
+ *
+ * @param {ConfigsOptions} options state decorator options
+ *
+ * @returns {Function} state decorator
+ */
+export function UserData(options: UserDataOptions) {
```
adrianschmidt, line 21: Is `options` optional? It doesn't look like it since there's no default value, but maybe it still is? I'm just asking since the only property, `options.name`, is optional, so it would make sense if the object is optional too…

jgroth: No it's not from the way it is declared right now, but maybe it should be :slightly_smiling_face:

packages/decorators/src/state/user-data.ts
```diff
@@ -0,0 +1,27 @@
...
+    PlatformServiceName,
+    StateOptions,
+} from '@limetech/lime-web-components-interfaces';
+import { createStateDecorator, StateDecoratorConfig } from '../factory';
+
+/**
+ * Options for user data
+ */
+export interface UserDataOptions extends StateOptions {
+    name?: string;
```
adrianschmidt, line 11: This is the key for the data, right? I think `key` would make it less ambiguous what this is.

Also, we could add a docblock for the property too… 🙂

jgroth: Yeah `key` might make more sense here, and we should probably add the docblock as well :+1:

packages/decorators/src/state/user-data.ts
adrianschmidt: Yeah, if no part of the config is actually required, it would make sense to be able to write:

```
@UserData()
@State()
private userData;
```

Rather than:

```
@UserData({})
@State()
private userData;
```

adrianschmidt (approved):

adrianschmidt (changes requested):
Just to prevent this getting merged before we're done with the other necessary PRs.

jgroth:
Lundalogik/lime-core#984 has been merged so I think we should be able to merge this one as well now :slightly_smiling_face:

adrianschmidt (approved):

adrianschmidt (changes requested):
Wait, is this breaking?

jgroth:
> Wait, is this breaking?

Hmm no it should not be, there is just new stuff added :thinking:

adrianschmidt (approved):
No, we're just adding a decorator, so not breaking, I was just confused 😅

PR #297, opened 2021-09-28
Refactor step 3

jgroth:
More continuation on #269 
This moves a few services to the new package. I also added some overloading of the `Platform.get()` method so you can get more info out of it when using the core services, e.g.

```ts
const http = platform.get('http');
```

Previously you would have to explicitly set the type of `http` to the http service, but now with the overloading TypeScript knows that it will be a `HttpClient`


packages/interfaces/src/limeobject.interface.ts
```diff
@@ -1,58 +1,16 @@
-import { Limetype, Property } from './limetype.interface';
+import { LimeObject as Base } from '../../lime-web-components/src';
+import { Property } from './limetype.interface';
 
 /**
- * Interface for the Lime object
+ * @deprecated has been renamed to better align with the Python class from
+ * `lime-core`. Use `LimeObject` from `@limetech/lime-web-components` instead.
```
Befkadu1, line 6: The name `LimeObject` in both Python class from `lime-core` and here is same.

jgroth: I don't think I understand what you mean, but this commit should already have been merged to `alpha` and should not be part of this PR :thinking:

Befkadu1: I meant it has not been renamed. 
Ok then how can we know which commits are part of this PR then?

jgroth: I will fix the commits :slightly_smiling_face:

jgroth: The commits are cleaned up now

Befkadu1 (commented):
It looks ok moving codes from one place to another. Can @adrianschmidt take a second look at this PR?

packages/lime-web-components/src/http/index.ts
```diff
@@ -4,7 +4,7 @@
  * By default, the service will work with the JSON data format. If anything but JSON is returned from the endpoint,
  * the `responseType` property in the `options` parameter needs to be set.
  */
-export interface HttpService {
+export interface HttpClient {
```
adrianschmidt: How is renaming an interface a refactor? 😬

Is this only used internally in lime-web-components? If not, wouldn't this be a breaking change?

packages/interfaces/src/service/commandbus-service.ts
```diff
@@ -0,0 +1,89 @@
+import {
```
adrianschmidt, line 1: `refactor(commandbus): deprecate old interfaces and export the new ones`

This should perhaps have been a feat or something… this seems like useful info to have in the changelog…

(Obviously, since this is just going into `alpha`, we can easily remedy that by just adding a commit with the scope and info we want to create the appropriate changelog entry, before we merge to `main`. Just commenting for the purpose if spreading my ideas and influencing future work 😁)

packages/interfaces/src/service/event-dispatcher-service.ts
```diff
@@ -0,0 +1,7 @@
+import { EventDispatcher } from '../../../lime-web-components/src/eventdispatcher';
```
adrianschmidt, line 1: Same comment for `refactor(eventdispatcher): deprecate old interface and export new one` of course 🙂

packages/interfaces/src/service/commandbus-service.ts
adrianschmidt: Oh, wait, this hasn't been merged yet 😆

Then we can adjust accordingly right here, if you agree 😛

packages/interfaces/src/service/http-service.interface.ts
```diff
@@ -0,0 +1,33 @@
...
+    HttpOptions as BaseOptions,
+    HttpParams as BaseParams,
+    HttpHeaders as BaseHeaders,
+    HttpResponseType as BaseResponseType,
+} from '../../../lime-web-components/src/http';
+
+/**
+ * @deprecated use `HttpClient` from `@limetech/lime-web-components` instead
+ */
+export interface HttpService extends HttpClient {}
```
adrianschmidt, line 12: Ah! Ok, I guess this answers how it's not a breaking change 😄

Clever!

packages/lime-web-components/src/http/index.ts
jgroth: Hmm I'm not sure what to call it instead of refactor :sweat_smile: Isn't renaming something a refactor?

It's not a breaking change since all interfaces from the current package still are exported exactly as before (but deprecated). If a consumer wants to implement the new interface instead they are free to do so, but the name is different and some other stuff might have changed as well (like some interfaces had optional methods that now are required and deprecated ones are gone)

adrianschmidt: Yeah, I realised later… I was reviewing commit by commit, which is why it looked very breaking at the time 😅

packages/interfaces/src/service/commandbus-service.ts
jgroth: Hmm yeah, do you think we need to list all these refactoring changes in the changelog? I was thinking of just making one entry for it or something when we are done explaining what has changed in more general terms. The moving/renaming of the interfaces are documented on the actual interface as well so you will see what has changed when you use the types inside your editor

packages/lime-web-components/src/http/index.ts
jgroth: Haha yeah :smile: 
I wanted to preserv as much of the commit history as possible so that's why there are so many commits in the PR (something I did not think about in the previous PRs unfortunately :disappointed: )

packages/interfaces/src/service/commandbus-service.ts
adrianschmidt: Yeah, making a single commit with a message about it is probably more reasonable. In fact, if we haven't made it possible to write longer changelog entries in commit messages except for breaking changes by the time we release this, it's probably best to just manually add a longer description of the changes to the changelog and the release description… 🤔

packages/lime-web-components/src/core/platform.ts
```diff
@@ -9,6 +11,7 @@ export interface LimeWebComponentPlatform {
      *
      * @param name the name of the service
      */
+    get(name: PlatformServiceName.Http | 'http'): HttpClient;
```
Befkadu1, line 18: Is this change should be in this commit?

adrianschmidt:
I know I've had at least a glance at this before, and I think it will just take more time than it's worth if I'm to really go through and understand all details of the code now, so I think we should merge this when @Befkadu1 is happy with it 👍

We'll want to do some thorough testing before merging alpha to main, so the last safety net will be there anyway 🙂

adrianschmidt:
FYI: The tests weren't running, so I did an empty `git commit --amend` and force-pushed, just to make them run.

PR #298, opened 2021-10-04
feat(hotkey-registry): create HotkeyRegistry interface

Befkadu1:
fix Lundalogik/crm-feature#2350


packages/interfaces/src/service/hotkey-registry.interface.ts
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Registers the command for a specific hotkey
+     *
+     * @param {string} hotkey the string representation of the hotkey
+     * @param {Object} command the command to trigger when the hotkey is pressed
+     *
+     * @example
+     * service.register('ctrl+s', mySaveCommand)
+     */
+    register: (hotkey: string, command: object) => void;
```
jgroth, line 16: Maybe a better name for these functions are `bind`/`unbind` and use the term `keybinding` instead of `hotkey`. At least for me those terms feels more common, but it might be because I'm a gamer :sweat_smile: Not sure if the same term is used in other applications

packages/interfaces/src/service/hotkey-registry.interface.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * Service for hotkey registry
```
jgroth, line 2: We need to become better at writing documentation for our services, how they are supposed to be used and how they work. Maybe move the examples below up here instead? We should also say something about the command bus and that commands needs to be registered first in order to be executed when a keybinding is pressed

packages/interfaces/src/service/hotkey-registry.interface.ts
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @private
+ */
+export interface HotkeyRegistry {
+    /**
+     * Registers the command for a specific hotkey
```
jgroth, line 8: How should we handle if a keybinding already exist? Should we allow it? Should we throw an exception? It seems VSCode allow multiple commands on the same keybinding but they also have different contexts connected to them.

keybindings

jgroth, line 8: How should we handle if a keybinding already exist? Should we allow it? Should we throw an exception? It seems VSCode allow multiple commands on the same keybinding but they also have different contexts connected to them.

keybindings

jgroth: This also brings up the question if we need to have contexts like this. If we do, how should we determine what they are?

packages/interfaces/src/service/hotkey-registry.interface.ts
```diff
@@ -0,0 +1,50 @@
...
+     * service.register('ctrl+s', mySaveCommand)
+     */
+    unregister: (hotkey: string, command: object) => void;
+
+    /**
+     * Get a list of CommandHotkeys
+     *
+     * @returns {CommandHotkey[]} the list of command and its specific hotkey
+     */
+    getCommandHotkeys: () => CommandHotkey[];
```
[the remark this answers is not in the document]
Befkadu1: `getRegisteredHotkeys` might be better

packages/interfaces/src/service/hotkey-registry.interface.ts
hannahu: When using keybindings in other apps like the gamer situations you mentioned, do they often use the same commands but in different contexts? Can we start with the simplest if we don't see the need at the moment and not allow multiple commands?

packages/interfaces/src/service/hotkey-registry.interface.ts
hannahu: I also like the bind/unbind and kebinding naming!

packages/interfaces/src/service/hotkey-registry.interface.ts
jgroth: No usually in games you can only have one command per keybind. If you add an already existing keybind, the old one will be unbound automatically

packages/interfaces/src/service/hotkey-registry.interface.ts
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Unregisters the command for a specific hotkey
+     *
+     * @param {string} hotkey the string representation of the hotkey
+     * @param {Object} command the command to trigger when the hotkey is pressed
+     *
+     * @example
+     * service.register('ctrl+s', mySaveCommand)
+     */
+    unregister: (hotkey: string, command: object) => void;
```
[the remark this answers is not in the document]
adrianschmidt: I can see one reason to require the command to deregister even if only one command can be registered to a given hotkey at a time, and that's to prevent a badly behaving component from deregistering another component's hotkey by mistake:

1. `quickComponent` successfully registers a command to the hotkey `s`
2. `badComponent` tries to register a command to `s`
3. `s` is already used, so the registration is rejected
4. `badComponent` ignores the rejection and goes on with its life
5. `badComponent` deregisters `s`, thus deregistering the command that `quickComponent` is still "using"

packages/interfaces/src/service/hotkey-registry.interface.ts
adrianschmidt, line 16: In `commandbus-service.ts` we use a type called `CommandIdentifier` for identifying commands. Could we use that here too, rather than requiring the command itself to be provided? I'm thinking that registering a command and registering the hotkey for a command isn't necessarily the same thing? There was mention of some problem with commands in menus, where the consumer registering the hotkey wasn't the one creating the command itself, if I understood correctly? Would only having to reference a command by its identifier help?

adrianschmidt, line 16: Rather than accepting a string that could contain any meta key, and normalising that string, like in the blog post, I think I'd prefer accepting the normalised value directly.

So, instead of accepting `ctrl+s`, what we accept is `{ ctrl: true, s: true }` (or perhaps `{ ctrl: true, key: 's' }`, if we think that would be better for us or more easily understood for the developer of the consumer).

That way, we can better control what people can input. We can strictly define what meta keys can be used (I'm sure we want to allow `shift`, not so sure about `ctrl`), and even which characters can be used, if we so wish.

packages/interfaces/src/service/hotkey-registry.interface.ts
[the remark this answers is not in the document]
adrianschmidt: Or… how about we return a de-registration function, like we do with observables?

packages/interfaces/src/service/hotkey-registry.interface.ts
jgroth: I think we need the actual command and not only the identifier when we bind something, otherwise we will need to create a new command when a keybinding is used and we do not have all necessary data to create a command inside this service. If we go with this approach we need to register command factories as well probably

Befkadu1: Wouldn't it cause the extra work in the component to find the command using `CommandIdentifier`?

packages/interfaces/src/service/hotkey-registry.interface.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * Service for hotkey registry
+ *
+ * @private
```
jgroth, line 4: We should add a note here about the service still being work in progress

packages/interfaces/src/service/hotkey-registry.interface.ts
jgroth: I think I like the string approach more. If I were a consumer I think I would prefer that since it's a lot less verbose. Also, from a configuration standpoint it will be easier to configure and then use directly without requiring a layer in between translating one format to another

packages/interfaces/src/service/hotkey-registry.interface.ts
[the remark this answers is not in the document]
jgroth: Both the dialog and event dispatcher services have pairs of `create`/`destroy` and `add`/`remove`.
I think the only place we return an unsubscribe function is from the `subscribe` function for the state services. The `subscribe` is often not used by the consumers and is more used by the state decorators. 

Returning a function will also require more bookkeeping for consumers since they will need to store the unsubscribe function somewhere.

packages/interfaces/src/service/hotkey-registry.interface.ts
adrianschmidt: I guess I'm not sufficiently familiar with how commands actually work to quite understand this… 😞

I was imagining something akin to events, where a command had a name, like "save" for example, and it could be triggered by telling something to "run the 'save' command"… but I guess it's not that simple then? 😕

jgroth: Yes that's true all commands have names like you say. However, we need the data associated with the command as well before we can run it and only the consumer knows what the data is.

adrianschmidt: I have to admit, arguments regarding how verbose something is makes me mentally check out. Am I really the only one who doesn't think how we choose to make the world work should be governed by how long something is to write?

The configuration argument, on the other hand, I think is relevant. But I don't understand it… Configure where, and do what conversion? If we're talking about configuring in Lime Admin, I think we should avoid looking too far in the future, although looking far enough ahead not to paint ourselves into a corner is obviously a good thing 😅. If we're talking about a config interface like that, I think it would be _much_ easier for the user to get presented with a list of commands that can take hotkeys, be allowed to enter a single character for each, and toggle checkboxes for each available modifier key, rather than having to write something like `ctrl+shift++`.

packages/interfaces/src/service/hotkey-registry.interface.ts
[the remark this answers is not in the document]
adrianschmidt: Ok. I don't have a strong opinion on either. Also, come to think of it, people tend to have a hard time understanding the `const unsubscribe = something.subscribe()` pattern when they first come across it, so it may well be for the best that we avoid that pattern for consumers… 👍

packages/interfaces/src/service/hotkey-registry.interface.ts
jgroth: > I have to admit, arguments regarding how verbose something is makes me mentally check out. Am I really the only one who doesn't think how we choose to make the world work should be governed by how long something is to write?

To me ease of use and readability are more important for an interface. Having a config object like you suggested feels more like an internal representation or implementation detail that we maybe do not want to expose to the consumer.

> Configure where, and do what conversion? 

Hmm I'm not sure what I meant haha :sweat_smile: I think I was multitasking and got confused by myself :upside_down_face: 

However, regarding having a config for this in Lime Admin I'm still not convinced we should not do this now already. If we skip doing this now, I think we will risk either having addons with hardcoded keybinds or having configs for keybinds spread out over multiple places in Lime Admin.

> I think it would be much easier for the user to get presented with a list of commands that can take hotkeys, be allowed to enter a single character for each, and toggle checkboxes for each available modifier key, rather than having to write something like ctrl+shift++.

To me the ideal way for a user to configure a keybind would be to select a command and press the actual keybind

packages/interfaces/src/service/keybinding-registry.interface.ts
```diff
@@ -0,0 +1,59 @@
...
+     * @param {Object} command the command to trigger when the keybinding is pressed
+     */
+    bind: (keybinding: string, command: object) => void;
+
+    /**
+     * Unbinds the command for a specific keybinding
+     *
+     * @param {string} keybinding the string representation of the keybinding
+     */
+    unbind: (keybinding: string) => void;
```
adrianschmidt, line 36: As long as this service is private, this is just fine. But if we release it with the limitation on one command per keybinding, and this signature for `unbind`, I'm betting money we'll eventually find this in an add-on:

```ts
keybinding.unbind('ctrl+s');
keybinding.bind('ctrl+s', mySaveFunction);
```

😄

packages/interfaces/src/service/hotkey-registry.interface.ts
Befkadu1: FYI when I added a note under the private, it created a warning
Screenshot 2021-10-05 at 14 15 39

packages/interfaces/src/service/keybinding-registry.interface.ts
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * The service is used to bind, unbind and get all keybindings.
+ *
+ * @example
+ * service.bind('ctrl+s', MySaveCommand)
+ *
+ * The above example is used to bind the command for a specific keybinding.
+ * The bind method takes the keybinding and command object as input. MySaveCommand
+ * is a command object which is registered using command bus. The command needs to
+ * be bound first in order to be executed when a keybinding is pressed.
```
adrianschmidt, line 12: > The command needs to be bound first in order to be executed when a keybinding is pressed.

My initial interpretation of this was that you have to bind the command using the commandbus service, before binding a keyboard shortcut, but when I then moved on to read the description for `bind()`, it sounded really confusing.

> Binds the command for a specific keybinding

So, the command has to be bound first, but this binds the command? I assumed that you were talking about two different binds, so I started to write this comment. Then I looked in commandbus-service.ts and realised it's called `register`, not `bind`, there. So now I'm back to thinking that "the command needs to be bound first in order to be executed when a keybinding is pressed" is referring to using the `bind` method here?

packages/interfaces/src/service/keybinding-registry.interface.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+/**
+ * Interface for command and its specific keybinding
+ */
+export interface Keybinding {
+    /**
+     * The string representation of the keybinding
+     */
+    keybinding: string;
```
hannahu, line 53: Maybe we should call this _keybind_ instead?

packages/interfaces/src/service/hotkey-registry.interface.ts
adrianschmidt: 🤷‍♂️

packages/interfaces/src/service/keybinding-registry.interface.ts
adrianschmidt: I disagree. A keybinding, many keybindings. In my opinion 😊

Befkadu1: VS uses this.

Screenshot 2021-10-05 at 14 32 49

packages/interfaces/src/service/keybinding-registry.interface.ts
Befkadu1: >  So now I'm back to thinking that "the command needs to be bound first in order to be executed when a keybinding is pressed" is referring to using the `bind` method here?

Yes

In the above doc I wrote like this above `...MySaveCommand is a command object which is registered using command bus`. I never used the word `bind` 😆

packages/interfaces/src/service/keybinding-registry.interface.ts
Befkadu1: Keybinding can be `ctrl+o`, `shift+0`, `a`, `b` etc.

hannahu: VS code seem to use _Keyboard Shortcut_s as the wrapping name.

adrianschmidt: Hmm… good point 👍

Maybe this should be called "keySequence" or just "key", or "keys" or something…

```
keybinding: Keybinding = {
	keys: 'ctrl-s',
	command: mySaveCommand,
};
```

I think "keybinding" fits for the interface (the outer object) here, because it's really only a _binding_ if it contains both the keys to be pressed and the command to execute, right?

packages/interfaces/src/service/keybinding-registry.interface.ts
adrianschmidt, line 12:
```suggestion
 * The above example is used to bind the command for a specific keybinding.
 * The bind method takes the keybinding and command object as input. MySaveCommand
 * is a command object which is registered using command bus.
```

packages/interfaces/src/service/keybinding-registry.interface.ts
Befkadu1: Aha now I understand what you were talking about. We have been discussing about the `Keybinding` interface name [here](https://github.com/Lundalogik/lime-web-components/pull/298#discussion_r721411568). You are welcome to suggest a better name for the interface.

packages/interfaces/src/service/keybinding-registry.interface.ts
```diff
@@ -0,0 +1,59 @@
+/**
+ * Service for Keybinding registry
+ *
+ * The service is used to bind, unbind and get all keybindings.
+ *
+ * @example
+ * service.bind('ctrl+s', MySaveCommand)
```
adrianschmidt, line 7:
```suggestion
 * service.bind('ctrl+s', mySaveCommand)
```

packages/interfaces/src/service/keybinding-registry.interface.ts
hannahu: I also think the inner should be renamed and keep the outer. I kinda like the _keySequence_ but maybe it's cleaner with _key/keys_.

Befkadu1: Then the inputs of `bind` and `unbind` methods to take key as a parameter then 
`bind: (key string, command: object) => void;`

hannahu: > Then the input of `bind` and `unbind` methods to take _key_ as a parameter instead of `keybinding` `bind: (key string, command: object) => void;`

Yes!

jgroth: I think `keySequence` is better, `ctrl+shift+c` or similar feels more like a sequence than a single key :slightly_smiling_face:

packages/interfaces/src/service/keybinding-registry.interface.ts
```diff
@@ -0,0 +1,57 @@
+/**
+ * Service for Keybinding registry
+ *
+ * The service is used to bind, unbind and get all keybindings.
+ *
+ * @example
+ * service.bind('ctrl+s', mySaveCommand)
+ *
+ * The above example is used to bind the command for a specific keys combination.
```
adrianschmidt, line 9:
```suggestion
 * The above example is used to bind the command for a specific key combination.
```

packages/interfaces/src/service/keybinding-registry.interface.ts
```diff
@@ -0,0 +1,57 @@
...
+ * service.unbind('ctrl+s')
+ *
+ * The above example shows how to unbind the command for a keybinding of 'ctrl+s'
+ *
+ * This service is work in progress
+ * @private
+ */
+export interface KeybindingRegistry {
+    /**
+     * Binds the command for a specific keys combination
```
adrianschmidt, line 22:
```suggestion
     * Binds the command for a specific key combination
```

packages/interfaces/src/service/keybinding-registry.interface.ts
```diff
@@ -0,0 +1,57 @@
...
+     *
+     * @param {string} keys the string representation of keys
+     * @param {Object} command the command to trigger when keys are pressed
+     */
+    bind: (keys: string, command: object) => void;
+
+    /**
+     * Unbinds the command for a specific keys combination
```
adrianschmidt, line 30:
```suggestion
     * Unbinds the command for a specific key combination
```

packages/interfaces/src/service/keybinding-registry.interface.ts
```diff
@@ -0,0 +1,57 @@
...
+     */
+    getKeybindings: () => Keybinding[];
+}
+
+/**
+ * Interface for command and its specific keybinding
+ */
+export interface Keybinding {
+    /**
+     * The string representation of keys combination
```
adrianschmidt, line 49:
```suggestion
     * The string representation of the key combination
```

packages/interfaces/src/service/keybinding-registry.interface.ts
```diff
@@ -0,0 +1,57 @@
...
+     */
+    keys: string;
+
+    /**
+     * The command to trigger when the keybinding is pressed
```
adrianschmidt, line 54:
```suggestion
     * The command to trigger when the key combination is pressed
```

packages/interfaces/src/service/keybinding-registry.interface.ts
```diff
@@ -0,0 +1,57 @@
...
+     *
+     * @param {string} keys the string representation of keys
+     * @param {Object} command the command to trigger when keys are pressed
+     */
+    bind: (keys: string, command: object) => void;
+
+    /**
+     * Unbinds the command for a specific keys combination
+     *
+     * @param {string} keys the string representation of keys
```
adrianschmidt, line 32:
```suggestion
     * @param {string} keys the string representation of a key combination
```

packages/interfaces/src/service/keybinding-registry.interface.ts
```diff
@@ -0,0 +1,57 @@
+/**
+ * Service for Keybinding registry
+ *
+ * The service is used to bind, unbind and get all keybindings.
```
hannahu, line 4: Can we make this general description more clear? Some words regarding what a keybinding really is like we've done in other parts of this file :)

packages/interfaces/src/service/keybinding-registry.interface.ts
```diff
@@ -0,0 +1,60 @@
...
+ * The above example is used to bind the command for a specific key combination.
+ * The bind method takes the key combination and command object as input.
+ * `mySaveCommand` is a command object which is registered using command bus.
+ * @example
+ * service.unbind('ctrl+s')
+ *
+ * The above example shows how to unbind the command for a keybinding of 'ctrl+s'
+ *
+ * This service is work in progress
+ * @private
```
adrianschmidt, line 21: Just pointing out: as long as this is private, we can make any changes we want without having to do a major version bump 👍🙂

PR #302, opened 2021-10-13
feat(create-dialog-command): add optional context property

FredrikWallstrom:


packages/commands/src/limeobject/create-dialog.command.ts
```diff
@@ -33,4 +33,9 @@ export class CreateLimeobjectDialogCommand {
      * Title of the dialog
      */
     public title?: string;
+
+    /*
+     * Current context
+     */
+    public context?: string;
```
jgroth, line 40: This should not be a string, right? 
```suggestion
    public context?: LimeWebComponentContext;
```

jgroth (changes requested):

packages/commands/src/limeobject/create-dialog.command.ts
FredrikWallstrom: 😮  Great catch!

jgroth (approved):

PR #307, opened 2021-11-04
update `hosted-git-info` to V2.8.9 in decorators package

Befkadu1:
fix Lundalogik/crm-feature#2420


packages/decorators/package-lock.json
```diff
@@ -2474,9 +2474,9 @@
       }
     },
     "node_modules/hosted-git-info": {
-      "version": "2.8.8",
-      "resolved": "https://registry.npmjs.org/hosted-git-info/-/hosted-git-info-2.8.8.tgz",
-      "integrity": "sha512-f/wzC2QaWBs7t9IYqB4T3sR1xviIViXJRJTWBlx2Gf3g0Xi5vI7Yy4koXQ1c9OYDGHN9sBy1DQ2AB8fqZBWhUg==",
+      "version": "2.8.9",
```
adrianschmidt, line 2477: Even if you want to keep these in separate commits, perhaps you can make a single PR with all the commits? That way we just have to review and test once, which would save a lot of time.

Befkadu1: ok, let this one go through and I will make one PR for others if it is ok

adrianschmidt (approved):

PR #310, opened 2021-12-14
feat(filter-service.interface): delete multiple filters at the same time

Befkadu1:
This change makes a possibility to delete multiple filters at the same time. The `delete` function of the `FilterService`interface takes either a single filter or a list of filters to delete.

fix Lundalogik/crm-feature#2555


packages/interfaces/src/service/state/filter-service.interface.ts
```diff
@@ -13,9 +13,9 @@ export interface FilterService extends StateService {
...
      * Deletes the filter
      *
-     * @param id a filter id to delete
+     * @param filters a list of filters to delete
      *
-     * @returns {Promise<void>} a promise that will be resolved when the filter is deleted
+     * @returns {Promise<void>} a promise that will be resolved when the filters are deleted
      */
-    delete?(filter: Filter): Promise<void>;
+    delete?(filters: Filter[]): Promise<void>;
```
adrianschmidt, line 20: While this is a breaking change for the services implementing the interface, we could allow supplying either a `Filter` or a `Filter[]`, so that the change doesn't become breaking for the consumers of the service… 🤔

What do you think @jgroth?

Befkadu1: Yea, you are right. we can allow either a Filter or a Filter[].

jgroth: Not sure if it is breaking if we allow both since the method is optional anyway :smile:

adrianschmidt: Hmm… 🤔

That's a good question… I would assume that if the function is there, it has to handle the types as specified, but I'm not sure.

jgroth: This
```ts
interface FilterService {
    delete?(filter: Filter | Filter[]): Promise<void>;
}
```
should be the same as writing this (using method overloading instead)
```ts
interface FilterService {
    delete?(filter: Filter): Promise<void>;
    delete?(filters: Filter[]): Promise<void>;
}
```
So I guess it will not be breaking since both are optional :smile:

adrianschmidt: But then the consumer would not know if they can send in a `Filter`, a `Filter[]`, or either, to any given implementation…? 🤔

I'll take some time from the hack-day to try it out!

jgroth: They can use either of them. You can still only have one implementation and you need to be able to handle both argument types. It's a good way writing it like that if you need to have different return values, e.g.
```ts
interface A {
    foo(a: string): string;
    foo(a: number): number;
}

interface B {
    foo(a: string | number): string | number;
}
```
The implementation for `A` and `B` will probably be the same, but with `A` the consumer know that if you call it with `string` you will get a `string` back. With `B` you don't know if you get a `string` or `number`

adrianschmidt: That's not quite what I meant 🙂

I've tested it now, and you are correct that neither the implementation nor the consumer sees this change as breaking. The implementation happily builds even though it only supports `Filter`, not `Filter[]`, just as you predicted.

The problem is that I've now changed the consumer as shown in the screenshot below, and that _also_ builds perfectly happily:

image

But when I try to use that code with the unmodified implementation, despite having no build errors anywhere, I predictably run into this problem:

image

So, while Typescript doesn't consider this breaking to implementations, I think we should still mark it as breaking, and write in the breaking change description that this requires any implementation to be updated, but won't be breaking to consumers of those implementations.

Befkadu1: Ofc you will get this error. You send an array of objects [Filter] here but the `delete` function we have in `limeclient.Js` accepts a `Filter`. 

You have to update the code in `limeclient.Js` [here](https://github.com/Lundalogik/limeclient.js/blob/b0ca9cd4795d34e912ba9664c788e6b0390336c3/src/filters/repository.ts#L117)  like I did in my other [PR](https://github.com/Lundalogik/limeclient.js/pull/132/files).

adrianschmidt: > Ofc you will get this error. You send an array of objects `[Filter]` here however the `delete` function in `FilterRepository` in `limeclient.js` accepts a `Filter` object. The `id` of the `filter` will be `undefined`. You have to update the code in `limeclient.Js` [here](https://github.com/Lundalogik/limeclient.js/blob/b0ca9cd4795d34e912ba9664c788e6b0390336c3/src/filters/repository.ts#L117) like I did in my other [PR](https://github.com/Lundalogik/limeclient.js/pull/132/files).

Uh… yes, that's exactly what I said 🤨 And since you have to update some code, it's a breaking change, which was my point.

jgroth: It's not breaking since you are not forced to update the code with the new interface, and since it's optional the consumer does not really know if the version of the webclient supports this interface or not. If the method would have a different name, e.g. `deleteAll(filters: Filter[])`, then the consumer could check if that method was available or not.

adrianschmidt: So I guess my suggestion is bad then. It's better to either just change to using `Filter[]` so it's a proper breaking change, or add a new optional method, so the consumer can at least check, right?

adrianschmidt: > It's not breaking since you are not forced to update the code with the new interface

I guess that depends on whether your definition of "breaking change" is limited to "things that break the build if you don't update the code" 🤷‍♂️

adrianschmidt: The original change was probably better than the one I suggested. Sorry to confuse things so much 🤦‍♂️

jgroth: I would prefer not to make a breaking change and I'm fine with either

```ts
interface FilterService {
    delete?(filter: Filter | Filter[]): Promise<void>;
}
```
or
```ts
interface FilterService {
    delete?(filter: Filter): Promise<void>;
    delete?(filters: Filter[]): Promise<void>;
}
```
Since they are optional anyway I don't think we need to mark it as breaking. I think maybe we just need to be better at documenting what version the webclient is supporting

adrianschmidt: I abstain my vote. Do as you prefer 😅

jgroth:
I suggest we close both this PR and the one in limeclient.js. We can implement this feature without making any changes to the interface. The implementation in limeclient.js right now is just looping over the filters and deleting them one by one. This can easily be achieved from the consumer side instead. 

It's a bit problematic to have the interface like this now. What should happen if one or more of the deletes fail? Since we don't have an endpoint for deleting multiple filters there's no easy way to rollback, and the way the interface is defined right now there's no way for the consumer to know which filters were deleted and which filters are still remaining. If the consumer would delete them one by one instead they are in control of what to do next.

Befkadu1:
> What should happen if one or more of the deletes fail?

Even if we implement this on consumer side the user still doesn't know which filters were deleted and which filters are still remaining when 1 or more deletes fail. 

For example you want to delete 3 filters (`Solna`, `Lund` and `Kalmar`), and 2  (`Solna` and `Lund`) of them have already deleted for some reason (may be by other user). You will get some errors. The user still doesn't know which filter causing the errors. The only thing he see is Error occurred as a notification. 

May be we don't need this feature for now then ???

Screenshot 2021-12-17 at 09 47 33

jgroth:
No that's true, but there's at least the possibility for the consumer to do something about it. That's why I think it should be done by the consumer instead since we cannot implement a proper solution for this right now as a generic service for anyone to use.

jgroth:
> May be we don't need this feature for now then ???

The feature itself has nothing to do with this discussion :slightly_smiling_face: We can still have it. Whether we implement it the same way in the consumer or not is another question, but I don't think we should implement some kind of rollback logic in the frontend right now

adrianschmidt:
> I suggest we close both this PR and the one in limeclient.js. We can implement this feature without making any changes to the interface. The implementation in limeclient.js right now is just looping over the filters and deleting them one by one. This can easily be achieved from the consumer side instead.
> 
> It's a bit problematic to have the interface like this now. What should happen if one or more of the deletes fail? Since we don't have an endpoint for deleting multiple filters there's no easy way to rollback, and the way the interface is defined right now there's no way for the consumer to know which filters were deleted and which filters are still remaining. If the consumer would delete them one by one instead they are in control of what to do next.

Very good points, and I totally agree!

@Befkadu1 Just to clarify, in case you don't know: When we say "consumer", we don't mean the end user. We mean the code using the code we're talking about. So, in this case, the "consumer" is the code that calls the `FilterService.delete()` function 🙂

Befkadu1:
> @Befkadu1 Just to clarify, in case you don't know: When we say "consumer", we don't mean the end user. We mean the code using the code we're talking about. So, in this case, the "consumer" is the code that calls the `FilterService.delete()` function 🙂

Yes I understood that. That is why I used the word ` consumer` in this case `crm-comp` and the `user` for example `Appers` or any admin that creates the filters. I think I know what I am writing about 😆

Befkadu1:
This PR is closed because it is enough to implement the future on the consumer side as we discussed above

adrianschmidt:
> > @Befkadu1 Just to clarify, in case you don't know: When we say "consumer", we don't mean the end user. We mean the code using the code we're talking about. So, in this case, the "consumer" is the code that calls the `FilterService.delete()` function 🙂
> 
> Yes I understood that. That is why I started my point ` Even if we implement this on consumer side ....` in this case `crm-comp` and the `user` for example `Appers` or any admin that creates the filters. I think I know what I am writing about 😆

👍 I just wasn't sure, so wanted to make sure we didn't misunderstand each other 🤗

PR #311, opened 2021-12-20
feat(interfaces): add interface `SaveLimeObjectError`

adrianschmidt:
Required by Lundalogik/lime-crm-components#705


packages/commands/src/limeobject/save-object.error.ts
```diff
@@ -0,0 +1,7 @@
+import { LimeObject } from '@limetech/lime-web-components-interfaces';
+
+export class SaveLimeObjectError extends Error {
+    constructor(public limeobject: LimeObject) {
+        super();
```
jgroth, line 5: I think we might want to set a message here as well, e.g.
```suggestion
        super(`Could not save limeobject ${limetype}:${id}`);
```

jgroth: Also, there might be some problems with having a class defined here and it might be better to just have an interface and implement it somewhere else. Having the class here will include it in every addon using `lime-web-components`, so there will be multiple instances of the `SaveLimeObjectError` class. We will throw one version of it from our implementation and the addon might catch it like this, but it will fail since the two classes are not the same even though they have the same name :disappointed: 

```ts
try {
    // save limeobject
} catch (e) {
    if (e instanceof SaveLimeObjectError) {
        // do something special
    }
}

```

adrianschmidt: Ah, good input! 👍

adrianschmidt: Do you think we should keep it here in `commands` though, or should I move it to `interfaces`?

jgroth: Hmm :thinking: Maybe `interfaces` makes more sense?
They will all be merged soon hopefully anyway :sweat_smile:

packages/commands/src/limeobject/save-object.error.ts
```diff
@@ -0,0 +1,5 @@
+import { LimeObject } from '@limetech/lime-web-components-interfaces';
+
+export interface ISaveLimeObjectError extends Error {
```
jgroth, line 3: I'm not a huge fan of Hungarian notation and we don't use it anywhere else so I think we should get rid of the `I` here. If it clashes somewhere else I would rather see we either rename the import on the consumer side or give the implementation class another name

adrianschmidt: Yeah, I really dislike Hungarian notation too (well, this kind, [this _other_ kind of Hungarian notation](https://www.joelonsoftware.com/2005/05/11/making-wrong-code-look-wrong/) actually looks really interesting and useful), but I had the problem that `export class SaveLimeObjectError implements SaveLimeObjectError` didn't work, and I forgot that I could rename the import. I'll absolutely rename the import instead. Thanks! 🙏

adrianschmidt: ⚡

packages/commands/src/limeobject/save-object.error.ts
adrianschmidt: ⚡ Moved to `interfaces`.

jensgustafsson (approved):

PR #313, opened 2022-01-19
feat(interfaces): adds session interface

hannahu:
fix Lundalogik/crm-feature#2611


packages/interfaces/src/session.interface.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Name of the application database
+     */
+    database: string;
+
+    /**
+     * Application's display name
+     */
+    display_name: string;
```
adrianschmidt, line 39: Could we use lowerCamelCase instead? 🙏

Same for `session_timeout` and `user_id`.

(Of course, if that makes things very messy or complicated somewhere else, it might make sense to let these be snake_case, but it does tend to be a little annoying in the frontend, when you're used to camelCase, and the linter might complain too…)

packages/interfaces/src/session.interface.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    expirationtime?: string;
```
adrianschmidt, line 41: This is the only property that doesn't have a docstring.

packages/interfaces/src/session.interface.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Current language
+     */
+    language: string;
+
+    /**
+     * Time of login
+     */
+    logintime: string;
```
adrianschmidt, line 61: It feels like we should be consistent and make these either lowerCamelCase (preferably), or snake_case:

`logintime` → `loginTime`/`login_time`
`logouttime` → `logoutTime`/`logout_time`
`expirationtime` → `expirationTime`/`expiration_time`

packages/interfaces/src/session.interface.ts
```diff
@@ -0,0 +1,82 @@
...
+     */
+    active: boolean;
+
+    /**
+     * Enabled addons
+     */
+    addons: Record<string, any>;
+
+    /**
+     * Wheter the user is an admin or not
```
adrianschmidt, line 14:
```suggestion
     * Whether the user is an admin or not
```

packages/interfaces/src/session.interface.ts
```diff
@@ -0,0 +1,82 @@
...
+     */
+    admin: boolean;
+
+    /**
+     * Apm config
+     */
+    apm: {
+        SERVICE_NAME: string;
+        SERVER_URL: string;
```
adrianschmidt, line 23: Is there a specific reason these two are uppercase? Are they constants in some way the other ones aren't?

packages/interfaces/src/session.interface.ts
```diff
@@ -0,0 +1,82 @@
...
+     */
+    addons: Record<string, any>;
+
+    /**
+     * Wheter the user is an admin or not
+     */
+    admin: boolean;
+
+    /**
+     * Apm config
```
adrianschmidt, line 19: Since I'm not 100 % sure what APM is, this is a little unhelpful to me. On the one hand, I'm guessing if I don't know what APM is, I maybe shouldn't touch it, but on the other hand, it feels like it wouldn't hurt to spell it out here.

Perhaps something like `APM (application permission management) config`? (Did I guess correctly for what the abbreviation means?)

adrianschmidt (commented):
Not to be considered a complete review, since I don't know what this is needed for, so I can't really say anything about any of the stuff that actually matters (architectural design stuff), but I had a quick glance and had some comments on less important details. Nevertheless I thought I'd post them.

packages/interfaces/src/session.interface.ts
jgroth: I agree. If we have a public interface intended for the frontend we should make sure all names are in camel case. 

Since there is no interface for this right now, it is technically not a breaking change. However, there might be a change in behavior if we rename the keys and have typed it to `any` all over the place. A solution to that could be to just leave all the snake case names on the object and also copy them to a camel case equivalent. That way we will get a warning if we use the interface somewhere and have used it incorrectly, but it would still work when typed as `any` :slightly_smiling_face:

packages/interfaces/src/session.interface.ts
jgroth: I think these are some config key/value pairs. We could just type it like this instead maybe
```ts
apm: Record<string, unknown>,
```

packages/interfaces/src/session.interface.ts
jgroth: It's something about logging in elastic if I remember correctly, but I don't know the details about how it works

packages/interfaces/src/session.interface.ts
```diff
@@ -0,0 +1,82 @@
+/* eslint-disable camelcase */
```
jgroth, line 1:
```suggestion
```

packages/interfaces/src/session.interface.ts
```diff
@@ -0,0 +1,78 @@
...
+     * Name of the application database
+     */
+    database: string;
+
+    /**
+     * Application's display name
+     */
+    displayName: string;
+
+    expirationTime?: string;
```
hannahu, line 37: Should we add a docstring for this property as well as Adrian also mentioned? :D

jgroth: I did not write anything because I was not sure what the value was :smile: 
It was always set to `null` when I tested it so I could not get any clues from that

hannahu: Hahah .D

PR #315, opened 2022-01-23
feat(interfaces): add timeofday as a property type

jensgustafsson:
fix: Lundalogik/crm-feature#2615

requires the latest lime-core release

You need to rebuild the metadatacache in order to test this. The easiest way to do so is to run the following SQL query in the database:

```
truncate table [metadatacache]
```

You also need to activate the feature switch `treat_time_as_time_of_day` in the system level config (`lime_config`)


adrianschmidt (approved):

adrianschmidt (approved):
Ah, yeah, OK, seems like this change was made in the right place. We apparently have a weird file structure in lime-webc-components-interfaces…

PR #317, opened 2022-02-02
feat(interfaces): adds user interface

hannahu:


packages/interfaces/src/user.interface.ts
```diff
@@ -0,0 +1,52 @@
...
+export interface UserGroup {
+    name: string;
+    id: number;
+    description: string;
+}
+
+interface UserCoworker {
+    id: number;
+    limetype: string;
+}
```
jgroth, line 52: These could use some doc blocks as well :slightly_smiling_face:

jgroth (commented):
Looks good! Some missing documentation and there are a few lint errors as well :slightly_smiling_face:

hannahu:
> Looks good! Some missing documentation and there are a few lint errors as well slightly_smiling_face

Oki!!

packages/interfaces/src/user.interface.ts
```diff
@@ -0,0 +1,52 @@
+export interface User {
+    /**
+     * The name of the user
+     */
+    username: string;
+
```
Befkadu1, line 6: do we need to have `isAdmin: boolean` or `admin: boolean` ?

hannahu: This is what we get from the endpoint, what you suggest doesn't exist in the user data :)

Befkadu1: Aha I confused it with Session

Befkadu1 (dismissed):

jgroth (approved):

PR #346, opened 2022-02-24
Create new navigator

civing:
fix Lundalogik/crm-feature#2688


jgroth (commented):
Looks good to me. We need to improve the documentation, so if we are not doing it in this PR we need to create a new issue to fix that

maltja:
> Looks good to me. We need to improve the documentation, so if we are not doing it in this PR we need to create a new issue to fix that

fixed 🙂

packages/interfaces/src/service/navigation-service.interface.ts
```diff
@@ -0,0 +1,13 @@
...
+ * @experimental
+ */
+export interface Navigator {
+    /**
+     * Navigate to a url
+     *
+     * @param {string} url the url string
+     */
+    navigate(url: string): void;
+}
```
jgroth, line 13: We should probably also add a type/interface for the event that gets dispatched from this service after navigating

maltja: fixed, added type Location. 🙂

jgroth (dismissed):
Looks good! I think we need to work on the documentation a bit, but let's create a new issue for that so we can get this merged :slightly_smiling_face:

jgroth (approved):

jgroth (approved):

PR #347, opened 2022-02-24
feat(router-registry-service): add new interface for router registry

civing:
fix Lundalogik/crm-feature#2690


packages/interfaces/src/service/route-service.interface.ts
```diff
@@ -2,6 +2,7 @@ import { Expression } from '../query.interface';
 
 /**
  * Service for routing
+ * @deprecated Has been replaced with a new routing service
```
jgroth, line 5:
```suggestion
 * @deprecated Has been replaced with the new `Navigator` service
```

jgroth (commented):
Looks good to me, but as I said on the other PR we need to improve the documentation or create a new issue to fix that :slightly_smiling_face:

packages/interfaces/src/service/route-registry-service.interface.ts
```diff
@@ -0,0 +1,40 @@
...
+     */
+    registerRoute(urlPattern: string, component: string): void;
+
+    /**
+     * Find the component for a given URL.
+     *
+     * @param {string} url - The URL to match against.
+     * @return {MatchedComponent | undefined} Returns the component along with route parameters if found.
+     */
+    findComponent(url: string): MatchedComponent | undefined;
```
jgroth, line 22: Should this be a URL or just the pathname? Did we decide to just use the pathname?

packages/interfaces/src/service/route-registry-service.interface.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+export interface RouteRegistry {
+    /**
+     * Register a route for a component.
+     *
+     * @param {string} urlPattern - URL pattern used to match a component.
+     * For more information about URL patterns read [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/API/URL_Pattern_API).
+     * @param {string} component - The component to register to the given URL pattern.
+     */
+    registerRoute(urlPattern: string, component: string): void;
```
jgroth, line 14: Kind of the same comment here, if we only require the pathname and if we recommend using that, maybe we should name this without including "URL" in the name

packages/interfaces/src/service/route-registry-service.interface.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * Matched path parameters.
+     */
+    props: Record<string, string>;
+
+    /**
+     * The extracted query parameters.
+     */
+    query: Record<string, string>;
```
jgroth, line 39: This should probably be `Record<string, unknown>` if we are going to try and parse the values

packages/interfaces/src/service/route-registry-service.interface.ts
```diff
@@ -0,0 +1,40 @@
...
+export type MatchedComponent = {
+    /**
+     * The name of the component.
+     */
+    name: string;
+
+    /**
+     * Matched path parameters.
+     */
+    props: Record<string, string>;
```
jgroth, line 34: Should we try and parse these as well? If so, we could change this to `Record<string, string | number>` or whatever makes more sense

jgroth (approved):
Looks good! Most of the comments are about improving the documentation and I think we can fix that in a separate issue so we can get this merged :slightly_smiling_face:

packages/interfaces/src/service/route-registry-service.interface.ts
civing: It makes sense to have them returned as numbers if they are numerics. 👍 We will fix this

packages/interfaces/src/service/route-registry-service.interface.ts
civing: Parse the values with JSON and treat them as unknown.

civing: (We will move the search params and put it in the Location instead)

packages/interfaces/src/service/route-registry-service.interface.ts
maltja: fixed!

packages/interfaces/src/service/route-registry-service.interface.ts
maltja: ⚡ changed it to pathPattern instead

packages/interfaces/src/service/route-registry-service.interface.ts
maltja: ⚡  changed to path

packages/interfaces/src/service/route-registry-service.interface.ts
maltja: ⚡

jgroth (approved):

jgroth (approved):

PR #364, opened 2022-03-01
Improve navigation docs

jgroth:


packages/interfaces/src/service/route-registry-service.interface.ts
```diff
@@ -1,23 +1,56 @@
...
+ * has been registered, it can be navigated to using the {@link Navigator}
+ * service.
+ *
+ * When registering a route, any parameters in the path will be used as props
+ * on the component when it is being rendered. Parameters will be of type
+ * `string`, when the parameter can be parsed as a number the type will instead
+ * be `number`.
+ *
+ * @note Special handling will be used for parameters named `limetype` and `id`
+ * to create a {@link LimeWebComponentContext} for the component
```
jgroth, line 18: We have not implemented it like this yet, but I think we should

packages/interfaces/src/service/navigation-service.interface.ts
```diff
@@ -1,16 +1,37 @@
...
  * @experimental
  */
 export interface Navigator {
     /**
-     * Navigate to a url
+     * Navigate to a new location
      *
-     * @emits navigate of type {@link NavigationEvent} using the event dispatcher when navigating to the new URL.
-     * @param {string} url the url string
+     * @param {string} path path to the location
```
vicgeralds, line 32: renaming url to path suggests passing a query string or hash isn't supported/recommended.

That's fine I think, as long as it will be clear how to pass additional location data.
(For example, `navigate(path, { query, hash, state })`)

Still, I think we should support navigating to a URL some way if we want to handle link clicks.

Also, maybe navigating without changing the path could be convenient, using `navigate('?query=value')` or `navigate('#fragment')`?

packages/interfaces/src/service/route-registry-service.interface.ts
```diff
@@ -1,23 +1,56 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { Navigator } from './navigation-service.interface';
```
vicgeralds, line 2: Can we use `import type` instead to silence eslint?

packages/interfaces/src/service/navigation-service.interface.ts
```diff
@@ -1,16 +1,37 @@
...
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { RouteRegistry } from './route-registry-service.interface';
+
 /**
- * Service for navigating to a url.
+ * The {@link Navigator} service lets you navigate to a new location within the
+ * application, or update the current location with new data. A location is
+ * defined by a path and optional data.
+ *
+ * When navigating to a new location, a new entry will be pushed to the history
```
vicgeralds, line 11: I'm reading this and wondering if replacing the location should emit a navigate event or not. But maybe that only needs to be clarified when we add a method for that.

packages/interfaces/src/service/navigation-service.interface.ts
```diff
@@ -1,16 +1,37 @@
...
+ * stack in the browser and the {@link EventDispatcherService} will emit a
+ * {@link NavigationEvent}.
```
vicgeralds, line 13: I think the name of the event emitted (`navigate`) should still be documented somewhere other than in the example.

packages/interfaces/src/service/navigation-service.interface.ts
jgroth: Yeah, the reason why I renamed it was because I think we should restrict it to only be the path. I don't think we should let the consumer handle the encoding of the query parameters and those should be passed as an object instead
Maybe we should add an overloaded method like this or something
```ts
    navigate(location: Location): void;
```
or
```ts
    navigate(path: string, params: Omit<Location, 'path'>): void;
```

packages/interfaces/src/service/route-registry-service.interface.ts
jgroth: Oh, I did not know you could write it like that. Unfortunately it did not seem to work :disappointed: I guess it's because they are only used in the docstrings

packages/interfaces/src/service/navigation-service.interface.ts
jgroth: Yeah I was a bit vague on purpose here since we have not fully decided how to implement it yet

packages/interfaces/src/service/navigation-service.interface.ts
vicgeralds: Yes, that could work. But (some of) the location properties should optional, right?

packages/interfaces/src/service/navigation-service.interface.ts
jgroth: I was not sure how best to document it. I wanted to move it from the method since it will also be emitted when hitting the back button etc, but maybe we can have it in both places. 
Is the `@emits` clear enough? I don't think it was clear that the event was named `navigate` but maybe that is just me :slightly_smiling_face:

packages/interfaces/src/service/navigation-service.interface.ts
jgroth: Yeah most of them should probably be optional :slightly_smiling_face:

packages/interfaces/src/service/navigation-service.interface.ts
vicgeralds: I understand. I think you could add  `@event navigate` when documenting `NavigationEvent` as well, but yes, it wasn't obvious how to document the event name using that for me either.

(We thought `@emits EventDispatcherService#navigate` looked even more strange.)

packages/interfaces/src/service/route-registry-service.interface.ts
vicgeralds: OK, too bad. I know you can import types in jsdoc as well, but that's even more ugly 😄 

```typescript
/** @typedef { import('./navigation-service.interface').Navigator } Navigator */
 ```

File: packages/interfaces/src/service/navigation-service.interface.ts (hunk 2 above)
Comment (vicgeralds) [reply packages/interfaces/src/service/navigation-service.interface.ts:32]: 👍  Let's do that in another PR then :)

File: packages/interfaces/src/service/route-registry-service.interface.ts (hunk 3 above)
Comment (jgroth) [reply packages/interfaces/src/service/route-registry-service.interface.ts:2]: Omg :dizzy_face: 
Maybe we can configure the linter and ignore it if `import type` is used, not sure if that makes sense or not

File: packages/interfaces/src/service/navigation-service.interface.ts (hunk 2 above)
Comment (jgroth) [reply packages/interfaces/src/service/navigation-service.interface.ts:32]: Yes, I wanted this to just be documentation related :slightly_smiling_face:

File: packages/interfaces/src/service/route-registry-service.interface.ts [hunk 6]
```diff
@@ -1,23 +1,56 @@
...
+ * application specified by either a path or a path pattern. Once a location
+ * has been registered, it can be navigated to using the {@link Navigator}
+ * service.
+ *
+ * When registering a route, any parameters in the path will be used as props
+ * on the component when it is being rendered. Parameters will be of type
+ * `string`, when the parameter can be parsed as a number the type will instead
+ * be `number`.
+ *
+ * @note Special handling will be used for parameters named `limetype` and `id`
```
civing, line 17: I think we should clearly state that `limetype` and `id` are reserved keywords and should not be used for other purposes than to create a LimeWebComponentContext

packages/interfaces/src/service/route-registry-service.interface.ts
civing: This will be implemented in Lundalogik/crm-feature#2706

vicgeralds (approved):
Looks good

packages/interfaces/src/service/route-registry-service.interface.ts
jgroth: Clarified this section a bit :slightly_smiling_face:

civing: It's very clear now, super 👍

civing (dismissed):

vicgeralds (approved):

PR #368, opened 2022-03-01
feat(product-analytics): add a service for product analytics

### LINKED ISSUE #371: Create product analytic service

maltja:
fix Lundalogik/crm-feature#2704
fix #371


packages/interfaces/src/service/product-analytics.interface.ts
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * @experimental
+ */
+export interface ProductAnalytics {
+    /**
+     * Track events
+     *
+     * @param {string} eventName name of the event. Limited to 255 characters.
+     * @param {object} config a JSON object, NSDictionary, or Map containing key-value
+     * pairs to be associated with an event. Keys and values must be a number or string
```
jgroth, line 13: I think we should rephrase this and make the interface a bit cleaner and only allow regular objects. I'm not sure we need to mention "JSON object, NSDictionary, or Map", is that coming from Heap? The interface needs to be generic and not Heap specific.

Also, why are `boolean`s not allowed? :thinking:

Befkadu1: Yes I will rephrase that.  Keys and values must be a number or string according to their docs [here](https://developers.heap.io/reference/track)

packages/interfaces/src/service/product-analytics.interface.ts
```diff
@@ -0,0 +1,16 @@
...
+     */
+    track(eventName: string, config: object): void;
```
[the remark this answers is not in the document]
Befkadu1: ⚡

packages/interfaces/src/service/product-analytics.interface.ts
jgroth: Have you tried with `boolean`s? 
And again, we should try and be generic and not too coupled to a specific tool like Heap in case we want to use something else in the future. I think `boolean`s make sense, and if Heap does not support it the Heap specific implementation should convert it to something else.

Befkadu1: It looks booleans are supported even if their doc exclude it

packages/interfaces/src/service/product-analytics.interface.ts
```diff
@@ -0,0 +1,16 @@
...
+ */
+export interface ProductAnalytics {
+    /**
+     * Track events
+     *
+     * @param {string} eventName name of the event.
+     * @param {Record<string, unknown>} config an object containing a key-value
+     * pairs to be associated with an event.
+     */
+    track(eventName: string, config: Record<string, unknown>): void;
```
maltja, line 15: Why do we have `unknown` here and not `any`? I'm not sure I know the difference. But if you have any then the user can input anything? 🤔 
```suggestion
    track(eventName: string, config: Record<string, any>): void;
```

maltja (commented):
Looks good, I just have a comment regarding the track function 🙂

packages/interfaces/src/service/product-analytics.interface.ts
Befkadu1: As I understand the `unknown` type is the type-safe counterpart of `any`. Anything is assignable to `unknown`, but `unknown` isn't assignable to anything.
For example

```
let age: number;
let userAge: unknown;

userAge = 'This is some age';
userAge = 20;

age = userAge;
````

The statement age=userAge gives an error. I would say unknown is a better choice. @jgroth

jgroth: `unknown` is safer to use than `any` because it enforces type checking
I found this article that probably explains it better than I do :slightly_smiling_face: 
https://dmitripavlutin.com/typescript-unknown-vs-any/#1-unknown-vs-any

jgroth: What happened to `string | number | boolean` that were there before? :slightly_smiling_face:

maltja: Thanks for the explanation, let's keep it as unknown 🙂

PR #369, opened 2022-03-03
Add navigate params

vicgeralds:


packages/interfaces/src/service/navigation-service.interface.ts
```diff
@@ -33,8 +33,17 @@ export interface Navigator {
      * {@link EventDispatcherService} will emit a {@link NavigationEvent}
      *
      * @param {string} path path to the location
+     * @param query query string parameters to append to the URL
      */
-    navigate(path: string): void;
+    navigate(path: string, query?: Location['query']): void;
+
+    /**
+     * Update the current location with new data
```
jgroth, line 41: I'm not sure about this description. If you provide a different path in the params, it will be a new location rather than updating the current one. What do you think?

packages/interfaces/src/service/navigation-service.interface.ts
```diff
@@ -33,8 +33,17 @@ export interface Navigator {
...
+     *
+     * @emits navigate - When the location has been changed the
+     * {@link EventDispatcherService} will emit a {@link NavigationEvent}
+     */
+    navigate(params: Partial<Location>): void;
```
jgroth, line 46: Perhaps name this `location`?
```suggestion
    navigate(location: Partial<Location>): void;
```

jgroth (commented):
Looks good, just some minor comments :slightly_smiling_face:

jgroth (approved):

packages/interfaces/src/service/navigation-service.interface.ts
vicgeralds: done

packages/interfaces/src/service/navigation-service.interface.ts
vicgeralds: ok if I rename it to `to`? 🙂

vicgeralds: never mind, I renamed to `location` like you suggested 🙂

jgroth (approved):

PR #377, opened 2022-03-21
Refactor interfaces

jgroth:
It might seem that there are a lot of changes here with 40 commits, 74 changed files, and +1400 changed lines of code. However, very little work has actually been done. The full diff on GitHub gives a lot of changes, but each commit seems to better reflect the actual changes that were made. I will try and describe the workflow I had to make it easier to review. Also, this is only merging to `alpha` so a thorough review is probably not necessary. When we have moved all code from the old structure we should try it out and fix any issues we find.

1. I made one commit per service that just moved the file for that service from the old `interfaces` package to the `lime-web-components` package. This is to preserve the git history for the file.
2. Made one new commit where fixes were made to the moved file. Fixes might have included fixing broken imports, minor documentation updates, or renaming services with weird names. E.g. anything ending with `Service` I just removed the suffix and most services ending with `State` were changed to end with `Repository` instead. The commit also re-creates the old file that was moved but just re-exports the moved interfaces from the new file but with the old name if it was changed. It also marks the old interface as deprecated.
3. I also added some module augmentation so that each service can add its service name to the `PlatformServiceName` enum and add an overload to the `platform.get` service. This will improve developer experience a lot since TypeScript will now know what type of service is returned depending on the name that is passed as an argument to the `get` method. This should hopefully also work for other packages later if they want to register their own services on the platform (once we also have support for exporting and importing types for lime packages)

Hopefully, this should conclude most of the refactoring of the `interfaces` sub-package (which is also the largest one).


packages/lime-web-components/src/core/platform.ts
```diff
@@ -33,38 +33,8 @@ export interface LimeWebComponentPlatform {
...
     Route = 'route',
-    Notification = 'notifications',
-    Query = 'query',
-    CommandBus = 'commandBus',
-    Dialog = 'dialog',
-    EventDispatcher = 'eventDispatcher',
 
-    LimetypesState = 'state.limetypes',
-    LimeobjectsState = 'state.limeobjects',
     ApplicationState = 'state.application',
```
adrianschmidt: So, we could remove all of these, except two? How come ApplicationState and Route are still needed? 🤔

packages/lime-web-components/src/commandbus/commandbus.ts
```diff
@@ -77,7 +77,7 @@ export interface CommandMiddleware {
  * Events dispatched by the commandbus event middleware
  */
 // eslint-disable-next-line no-shadow
-export enum CommandEvent {
+export enum CommandEventName {
```
adrianschmidt, line 80: It seems a little mysterious to me that this is just renamed on its own, completely in isolation? If this is used internally, shouldn't there be some corresponding change where it's used? If it's used externally, shouldn't this be a breaking change? If it's not used at all… well… 😄

packages/interfaces/src/session.interface.ts
```diff
@@ -0,0 +1,6 @@
+import { Session as BaseSession } from '../../lime-web-components/src';
```
adrianschmidt, line 1: It's great that you took care to preserve git history by moving the file in a separate commit 👍

I think these "second commits" could be typed `feat` instead though, and say that the interface has been moved or something. So we get some mention of it in the changelog. What do you think?

adrianschmidt (commented):
Looks very good overall! I just had a few minor comments and questions.

packages/lime-web-components/src/dialog/dialog.ts
```diff
@@ -32,10 +32,9 @@ export interface DialogService {
...
-}
+export type DialogProperties = Record<string, unknown>;
 
-export interface DialogListeners {
-    [listener: string]: (event?: CustomEvent<any>) => void;
-}
+export type DialogListeners = Record<
+    string,
+    (event?: CustomEvent<any>) => void
+>;
```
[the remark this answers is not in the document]
jgroth: No, I realized now when I started to do some basic testing that they are not :smile: 
The old one still needs to be `any`, but the new one we probably want to redefine as `unknown` I will fix it :slightly_smiling_face:

packages/interfaces/src/service/task-service.interface.ts
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * @deprecated moved to `@limetech/lime-web-components`
+ */
+export interface TaskService extends TaskRepository {}
+
+/**
+ * @deprecated moved to `@limetech/lime-web-components`
+ */
+export enum TaskState {
```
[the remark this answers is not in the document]
jgroth: No, unfortunately, I found no way to make this work with `enum`s, so I had to duplicate them :disappointed: 

> Although, then perhaps we can't add the documentation that this one's deprecated?

Yes, that's the reason to why that solution will not work either :disappointed: 

Fortunately we only had a handful of `enum`s so there's not that much duplication

packages/lime-web-components/src/core/platform.ts
jgroth: `Route` is just deprecated since we have the new `Navigator` interface, so I did not bother to move that interface at all to the new package. The service will still exist on the platform though so they will be able to use it for the time being

The service for `ApplicationState` does not have any own methods and just implements the base interface for all the state services. I could still move it to it's own folder I guess, since the decorators corresponding to this service now lives in the `core` package in the next PR #381. I will look into it :slightly_smiling_face:

packages/lime-web-components/src/commandbus/commandbus.ts
jgroth: The old `enum` still has the old name. I wanted to rename the new one so that we later can create a type named `CommandEvent`, similar to how we did here https://github.com/Lundalogik/lime-web-components/blob/c2e310e0537340bbf8e94c4db8f43ba2f2e50cac/packages/interfaces/src/service/navigation-service.interface.ts#L87

packages/interfaces/src/session.interface.ts
jgroth: Oh yeah, sure we can do it like that :+1: :slightly_smiling_face:

packages/lime-web-components/src/commandbus/commandbus.ts
adrianschmidt: Ah! Then I understand! 👍

packages/lime-web-components/src/core/platform.ts
adrianschmidt: > `Route` is just deprecated since we have the new `Navigator` interface, so I did not bother to move that interface at all to the new package. The service will still exist on the platform though so they will be able to use it for the time being

Ah, that sounds very reasonable 👍

> The service for `ApplicationState` does not have any own methods and just implements the base interface for all the state services. I could still move it to it's own folder I guess, since the decorators corresponding to this service now lives in the `core` package in the next PR #381. I will look into it 🙂

Do what you feel is best. If it feels cleaner and more consistent to move it, and it's easy to do, then that might make sense. If it's complicated and/or doesn't really bring any benefit, I don't mind it being left where it is. I just asked to make sure there was _some_ reason for these two to be different from the rest 🙂

jgroth: > Ah, I see you've already moved it. Nice!

I haven't done anything yet! :smile:

adrianschmidt: 🤔

Maybe I was confused… hang on 😅

adrianschmidt: Ah, no I was confused…

I saw that the code I started the thread on was marked outdated, so I looked at the diff, and saw this:

image

And I guess my somewhat post-lunch-comatic brain didn't care to consider the difference between red and green lines 😂

jgroth:
I have squashed the fixups and fixed the commit messages :slightly_smiling_face:

packages/interfaces/src/service/dialog-service.interface.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * @deprecated moved to `@limetech/lime-web-components`
+ */
+export interface DialogService extends DialogRenderer {}
+
+/**
+ * @deprecated moved to `@limetech/lime-web-components`
+ */
+export interface DialogProperties extends BaseProperties {
+    [prop: string]: any;
```
adrianschmidt, line 16: This looks like it fixes the breaking change for `DialogProperties`, so that's ✅

packages/interfaces/src/service/dialog-service.interface.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+/**
+ * @deprecated moved to `@limetech/lime-web-components`
+ */
+export interface DialogListeners extends BaseListeners {}
```
adrianschmidt, line 22: But this one seems to just re-export the new `DialogListeners` that has different typing from the old one, so isn't this still breaking? Or does the re-export lose or reset the types somehow?

jgroth: No, they are the same. It's just another way of writing it :slightly_smiling_face:

adrianschmidt: Aha! Ok then 👍 🙂

packages/interfaces/src/service/state-service.interface.ts
```diff
@@ -0,0 +1,27 @@
+import {
```
adrianschmidt, line 1: > feat(state): deprecamove interface to @limetech/lime-web-components

😆 Good word!

I'll change it to match the other commits though 😛

adrianschmidt (approved):

PR #380, opened 2022-03-23
fix(interfaces): move product analytics interface to its own package

Befkadu1:
Product analytics will be moved to [Lundalogik/limepkg-product-analytics](https://github.com/Lundalogik/limepkg-product-analytics)


packages/interfaces/src/lime-web-component-platform.interface.ts
```diff
@@ -58,11 +58,6 @@ export enum PlatformServiceName {
      * @private
      */
     KeybindingRegistry = 'keybindingRegistry',
-    /**
-     * @note Work in progress, do not use!
-     * @private
```
adrianschmidt, line 63: Nice to see this was `private`! If it hadn't been, this would have been a breaking change! Good work! 👍

PR #381, opened 2022-03-23
Refactor decorators

jgroth:
This deprecates all code in the `decorators` package and moves it to the `lime-web-components` package. GitHub does not give a good preview of the changes here, but I followed the same pattern as in #377 

1. One commit that moves the file to the new location
2. A new commit that fixes any broken imports, documentation, and naming issues. Old decorator is deprecated and re-exports the moved code under the old name.

The most notable change in this PR is a new naming convention for the decorators. Previously we had issues with name collision between types from `interfaces` and `decorators`, e.g. we both had a type and a decorator named `Session` so it was difficult to do something like the following without re-naming the types in your own code.

```ts
class MyComponent {
    @State()
    @Session() // The session decorator
    private session: Session; // The type for the session (name collision)
}    
```

With this PR all decorators are now named `Select...`, e.g. `@SelectSession()` and `@SelectCurrentLimeType()`. This was taken from Redux which has the selector concept[^1] of selecting data from the redux store. The `NgRedux` package that we have used in the webclient also has something similar with the generic `@Select` decorator that allows you to select anything from the state so I think this naming convention makes sense.

[^1]: https://redux.js.org/usage/deriving-data-selectors


packages/lime-web-components/src/core/application.ts
```diff
@@ -10,7 +10,9 @@ import {
  * @param {StateOptions} [options] options for the state selector
  * @returns {Function} state decorator
  */
-export function SelectApplicationName(options: StateOptions = {}) {
+export function SelectApplicationName(
+    options: StateOptions = {}
+): PropertyDecorator {
```
adrianschmidt: Looks like this change belongs in the preceding commit?

packages/decorators/src/index.ts
```diff
@@ -7,4 +7,9 @@ export * from './state/limeobjects';
 export * from './state/limetypes';
```
adrianschmidt, line 7: Same thing here as in #377, regarding the commit messages, and making sure the changes appear in the changelog.

adrianschmidt (changes requested):
Two tiny things, otherwise it looks great 👍

packages/lime-web-components/src/core/application.ts
jgroth: It took a few tries but I think I managed to sort it out :smile:

packages/decorators/src/index.ts
jgroth: Fixed :slightly_smiling_face:

adrianschmidt (approved):
Looks good to me 👍

PR #386, opened 2022-03-24
Refactor commands

jgroth:
Moves all commands from the `commands` package to `lime-web-components`
Based on #381


adrianschmidt (approved):
Looks really straightforward to me 👍

Issue #405, opened 2022-04-12
Decorators: async `componentWillLoad` does not wait for async `connectedCallback` as expected, which leads to crashes


adrianschmidt:
This might be a problem in Stencil but the docs are a bit unclear.
Bug reported https://github.com/ionic-team/stencil/issues/3325

We should update our factory `createComponentWillLoad` to make sure it either waits for `connectedCallback`, or at least `await`s `ensureLimeProps`.

PR #406, opened 2022-04-12
fix(decorators): make sure `componentWillLoad` waits for decorators to load

adrianschmidt:
fix: #405


packages/decorators/src/factory.ts
```diff
@@ -188,6 +188,8 @@ function createComponentWillLoad(original: Function, properties: Property[]) {
     return async function (...args) {
         if (connectedComponents.get(this) === true) {
```
adrianschmidt, line 189: @jgroth I don't think this will ever be falsy, because on line 143, we create a `connectedCallback` regardless of whether the component has one already. That's going to be called before our `componentWillLoad` callback, and it runs `connectedComponents.set(this, true);` on line 164, before it "becomes" async by using `await` on line 166.

adrianschmidt, line 189: I think we should refactor this file a little, to make it easier to understand. Use some named functions and such.

Once I got my head around what this code is doing, I realised it's not very complicated at all, but it took me a good long while to get there. I think we can make it a lot easier for the next reader 😅

packages/decorators/src/factory.ts
```diff
@@ -188,6 +188,8 @@ function createComponentWillLoad(original: Function, properties: Property[]) {
...
             if (original) {
+                await ensureLimeProps(this);
```
Befkadu1, line 191: Did you test it? When the related view (omg I have to start using the name `side panel `) opens, its size is big at the  beginning

adrianschmidt: I might not have tested it with all the cases we had. I'll have a look.

(One problem I had when trying to test this is that when I run with locally installed dependencies, the bugs are less likely to appear to begin with, so it's hard to know if a bug disappeared because it's fixed or because I'm running a local dependency… 😞)

packages/decorators/src/factory.ts
jgroth: Aha, yeah that might be true :thinking: 
Not sure what the thought was here :smile:

jgroth: It was even more complicated and hard to understand before :smile: But yeah, we can probably make it even better :slightly_smiling_face:

packages/decorators/src/factory.ts
adrianschmidt: Hmm… you're right. The problem with loading the object explorer (global table) the second time has gone away, but the incorrect width of the side panel is still there…

I'm sure I saw it was gone at one point though… 🤔

jgroth: We have some unit tests for this file so we might want to write a test for this :slightly_smiling_face: 
We can probably copy the test on line 116 (or refactor and make it work with `componentWillLoad` as well)

packages/decorators/src/factory.ts
adrianschmidt: The code might have evolved a bit, and something that was once true isn't anymore 🙂

Or we just thought the code did something different from what it actually does. Both seem plausible enough 😄

packages/decorators/src/factory.ts
Befkadu1: What I do is, I just run `npm run build` then I will have a production build. `original` becomes undefined sometimes, then we can't set the `platform` and `context`. What if we set `await ensureLimeProps(this);` above the `if` condition (above line 190)?

adrianschmidt: Yeah, I run production builds too. With dev-builds, I don't seem to be able to recreate any of the bugs at all… so annoying 😣

adrianschmidt: Changing this from

```
if (original) {
    await ensureLimeProps(this);
    …
```

to

```
await ensureLimeProps(this);
if (original) {
    …
```

does solve the problem. But I don't think it's the correct solution, because in that case, there is no `componentWillLoad` callback defined by the component itself, so there is nothing that should have to wait for `ensureLimeProps`. So I think there is something _else_ that just happens to work because we're introducing a small delay. But introducing that small delay to _every single component_ that doesn't have a `componentWillLoad`, just to solve a timing issue with one specific component, isn't ideal… 😕

Befkadu1: We don't have a watch mode test in `lime-web-components`? hmm...

adrianschmidt: > We don't have a watch mode test in `lime-web-components`? hmm...

There is in the respective packages. So you can go to `…/packages/decorators/` and run `npm run test:watch`.

adrianschmidt: > Did you test it? When the related view (omg I have to start using the name `side panel `) opens, its size is big at the beginning

I added [this PR](https://github.com/Lundalogik/lime-crm-components/pull/971) for the sidepanel. That achieves the same delay, which fixes the size and animation, but doesn't add any delay for any components other than sidepanel 👍

jgroth (approved):

adrianschmidt:
@jgroth Squashed and rebased!

Befkadu1 (approved):

Befkadu1:
I found out that there are some components that don't have `componentWillLoad` method. In those components, it crashes. For example for belongs to filters like in `in-filter.tsx`, it crushes. So the question is should we need to add `componentWillLoad` method in many components or better to put `await ensureLimeProps(this);` above [this line](https://github.com/Lundalogik/lime-web-components/blob/dfb5ac043d85d187580a582f654d9786eb946c84/packages/decorators/src/factory.ts#L190)

adrianschmidt:
I'll have a look 👍

packages/decorators/src/factory.ts
adrianschmidt: After some deliberation and a bit of digging, we settled on #408, as suggested by @Befkadu1, instead of the PR linked above.

PR #412, opened 2022-04-14
Move decorator location

adrianschmidt:
Should not be rebased into `alpha`. I'll bring this into `alpha` manually once approved.

The review is regarding the commits _after_ the merge.


packages/lime-web-components/src/index.ts
```diff
@@ -17,3 +17,4 @@ export * from './device';
 export * from './filter';
 export * from './userdata';
 export * from './application';
+export * from './location';
```
jgroth, line 20: I think this should live together with the other navigation stuff in the `navigator` folder

packages/lime-web-components/src/location/decorator.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+/**
+ * Gets a named query param from the location
+ *
+ * @experimental
+ * @param {StateOptions} options state decorator options
+ * @returns {PropertyDecorator} state decorator
+ */
+export function QueryParam(options: LocationOptions): PropertyDecorator {
```
jgroth, line 19: For consistency, I think we should rename this to `SelectQueryParam`

packages/lime-web-components/src/index.ts
adrianschmidt: ⚡

packages/lime-web-components/src/location/decorator.ts
adrianschmidt: ⚡

I also renamed `LocationOptions` to `SelectQueryParamOptions` in a separate fixup.

adrianschmidt:
Superseded by #434

PR #413, opened 2022-04-14
feat(navigation): add interface to block navigation

civing:
Add methods to add and remove blockers for navigation. These can be used in the case that navigation
should be blocked, for instance when the user has unsaved changes that would be lost in the case of
navigation. The added blocker method should return either true or false depending on if the navigation
should be blocked or not.

fix: https://github.com/Lundalogik/crm-feature/issues/2767


packages/interfaces/src/service/navigation-service.interface.ts
```diff
@@ -56,6 +56,28 @@ export interface Navigator {
...
+     * If the blocker function returns true navigation is blocked. If it returns
+     * false navigation will proceed as normal.
+     *
+     * A location is provided to the blocker function. This could be used to later
+     * on resume the navigation, for instance as the result of some user response or
+     * equivalent.
+     *
+     * @param blocker function to be called before navigation
+     */
+    addBlocker(blocker: Blocker): void;
```
adrianschmidt, line 73: Have you thought of any way to distinguish which blockers to run and which not to? Like in the case of having one object in full mode, and another in the side-panel, and closing the side-panel, we need to block if the object in the side-panel has unsaved changes, but we don't need to block for unsaved changes in the main view.

I can see two general ways that can be achieved:

- Supplying some kind of "identifying" information with the registration of the blocker, so that the navigation service knows which blockers to run
- Calling all registered blockers upon navigation, and supplying some information (like the url we're about to navigate to, for example), that the blockers can use to determine if "its" unsaved changes are about to be wiped by the navigation, if not, it doesn't need to block

During standup, we also discussed whether opening and closing the side-panel should be a navigation event at all. If it's not, we _might_ not need this ability to know whether certain content will or will not be removed by a given navigation event. But I think it's likely we will want it, because:

- While closing the side-panel by clicking the x doesn't feel very "navigation-y", consider opening the side-panel instead; the user clicks a link (hopefully a proper html `<a href="…">` some time in the future) and we pick up that event. On a phone or similarly small window, we navigate to the object, because the window is too small to fit a side-panel, if the window is large enough, we open the side-panel. So already there we need to involve navigation in _some_ way, or at least _potentially_ involve navigation.
- And then there may already be an object open in an existing side-panel, in which case we need to check if we can safely close that, before opening another object there instead.
- Even if we allow stacked side-panels, we either need to handle unsaved changes when a panel becomes "not the front panel", or if a stack of side-panels gets closed, and there are unsaved changes in any of them.
- There's some amount of complexity here, so do we really want to create a parallell system of blocking, that does very similar things to the "main" navigation and blocking system?

packages/interfaces/src/service/navigation-service.interface.ts
```diff
@@ -56,6 +56,28 @@ export interface Navigator {
...
+    /**
+     * Add a blocker that can prevent navigation away from the current page, for
+     * example when there is some unsaved data that would otherwise be discarded.
+     *
+     * If the blocker function returns true navigation is blocked. If it returns
+     * false navigation will proceed as normal.
+     *
+     * A location is provided to the blocker function. This could be used to later
+     * on resume the navigation, for instance as the result of some user response or
+     * equivalent.
```
vicgeralds, line 69: I think this piece of information is more helpful if it's turned around like, "To implement [functionality], you can use this thing" rather than "There is this thing. You can use is it to implement [functionality]."

(also, "the equivalent of some user response" is just wordiness to me)

```suggestion
     * To later on resume navigation that was blocked, for instance as the result of some user response,
     * you may save the new location provided to the blocker function.
```

... but now that I think about it, how's the blocker function supposed to save the blocked location? Maybe it's more helpful if the navigator service could provide the most recently blocked location instead?

vicgeralds (commented):
A few suggestions for improvement

packages/interfaces/src/service/navigation-service.interface.ts
vicgeralds: I think the way we have discussed this is that all blockers should run, and any blocker can block the current navigation event. Each blocker should have the information needed to determine if we are about to navigate away from its "guarded" location.

Yes, currently we are sometimes opening/closing the sidepanel without navigating, so it's a bit tempting to make the sidepanel able to block some other way. We could possibly do this by watching its context, but as you point out, there are pitfalls and added complexity that we probably don't want to deal with. The good thing would be that we could have better separation of concerns and less duplication (a single blocker in the sidepanel, instead of one blocker for each route where the sidepanel is used). We will still need navigation-based blocking anyway, so I think that's the way to go to keep it simple.

packages/interfaces/src/service/navigation-service.interface.ts
adrianschmidt: Now, I haven't looked at the whole PR, so it's likely I'm missing stuff here, but:

It sounds a little odd to me that the blocking-function would be involved in resuming navigation… 🤔

Is the plan to let each blocking-function display its own warning dialog to the user? Unless we're sure we'll never need to have more than one blocking function registered at any one time, it seems to me like we should rather let some central thing control that. But I might be biased by our old solution… 🤷‍♂️ 😅

packages/interfaces/src/service/navigation-service.interface.ts
civing: > Have you thought of any way to distinguish which blockers to run and which not to? Like in the case of having one object in full mode, and another in the side-panel, and closing the side-panel, we need to block if the object in the side-panel has unsaved changes, but we don't need to block for unsaved changes in the main view.

The aim here has been to keep the blockers as simple and stupid as possible. So everyone adding blockers needs to control the logic whether to block navigation or not *and* make sure to remove the blockers when not in use. Like in your case, either that component adds one blocker at some centralised place and do the distinction there *or* the two different blockers are added at two separate places. Due to the unanimous voter pattern it's enough for one of them to say no to block navigation. Then of course those two components needs to sort out what dialogs to show and how, but this logic can't seep into the navigator domain.

> I can see two general ways that can be achieved:
> ...
> * Calling all registered blockers upon navigation, and supplying some information (like the url we're about to navigate to, for example), that the blockers can use to determine if "its" unsaved changes are about to be wiped by the navigation, if not, it doesn't need to block

We are suppling `Location` as input to each blocker function in the event of navigation

packages/interfaces/src/service/navigation-service.interface.ts
vicgeralds: Just realised my suggestion made this ambiguity worse:

"navigation that was blocked, [...] as the result of some user response"

packages/interfaces/src/service/navigation-service.interface.ts
adrianschmidt: > > Have you thought of any way to distinguish which blockers to run and which not to? Like in the case of having one object in full mode, and another in the side-panel, and closing the side-panel, we need to block if the object in the side-panel has unsaved changes, but we don't need to block for unsaved changes in the main view.
> 
> The aim here has been to keep the blockers as simple and stupid as possible. So everyone adding blockers needs to control the logic whether to block navigation or not _and_ make sure to remove the blockers when not in use. Like in your case, either that component adds one blocker at some centralised place and do the distinction there _or_ the two different blockers are added at two separate places. Due to the unanimous voter pattern it's enough for one of them to say no to block navigation. Then of course those two components needs to sort out what dialogs to show and how, but this logic can't seep into the navigator domain.

Yeah, that part isn't in question really, not by me at least. The problem is rather "who gets to vote"? If we have `blocker A` guarding `content A`, and `blocker B` guarding `content B`, and get a navigation event that will only remove `content B`, then `blocker A` should not get a vote, because `content A` is unaffected by the navigation. (Or, in the case where all blockers are run, I guess the analogy is more like: "How does `blocker A` know that it doesn't have to vote against the navigation event, even if the content it's guarding has unsaved changes?")

> > I can see two general ways that can be achieved:
> > ...
> > 
> > * Calling all registered blockers upon navigation, and supplying some information (like the url we're about to navigate to, for example), that the blockers can use to determine if "its" unsaved changes are about to be wiped by the navigation, if not, it doesn't need to block
> 
> We are suppling `Location` as input to each blocker function in the event of navigation

👍 Considering that we can figure out what to render based on a url, I guess we should also be able to figure out if a given component will _still_ be rendered after changing to the new url, so that ought to do it 🙂

packages/interfaces/src/service/navigation-service.interface.ts
```diff
@@ -85,3 +107,11 @@ export type Location = {
  * location is updated
  */
 export type NavigationEvent = CustomEvent<Location>;
+
+/**
+ * Function to determine if navigation should be blocked or not.
+ *
+ * @param location the intended location of the navigation
```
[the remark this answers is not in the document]
civing: Ah, you would like to also supply the old location? Hmm, yeah why not I guess? On the other hand a consumer could use the `getLocation` from the same API if needed?

packages/interfaces/src/service/navigation-service.interface.ts
civing: I added your suggestion for improvement to the doctoring in a fixup commit. Could you please have a look?

packages/interfaces/src/service/navigation-service.interface.ts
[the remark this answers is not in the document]
jgroth: I think the new location should be enough, since you can get the current one anyway :slightly_smiling_face:

packages/interfaces/src/service/navigation-service.interface.ts
jgroth: > At first I made the blocking functions async with the notion that the blocker should return true or false based on user input. But after talking to @jgroth it's clear that you shouldn't wait for something

Having them return a `Promise<boolean>` could be a valid use case, but I'm not sure how everything should work then :thinking: What would happen after all promises have been resolved? Should we retry the transition when all of them resolved to `false` perhaps? :thinking: 
Some of the old routes in the Angular router have these "Guards" and those could return a whole bunch of different things, `Promise<boolean>` being one of them

https://github.com/Lundalogik/lime-webclient/blob/main/frontend/webclient/src/app/guards/ngu-admin-guard.service.ts#L10

Having these return a promise seems to complicate the use case a bit, but we can always add it later if we think it seems like a good idea :slightly_smiling_face:

packages/interfaces/src/service/navigation-service.interface.ts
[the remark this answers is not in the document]
vicgeralds: You can of course get the current one, but if most blockers will need to compare it against the new location that could be annoying (for example, checking if `location.query.selected` is different).

Do we want a nice interface for blocking location _changes_ or a minimal one for blocking navigation events?

Compare https://stenciljs.com/docs/reactive-data#watch-decorator

packages/interfaces/src/service/navigation-service.interface.ts
```diff
@@ -64,9 +64,8 @@ export interface Navigator {
      * If the blocker function returns true navigation is blocked. If it returns
      * false navigation will proceed as normal.
      *
-     * A location is provided to the blocker function. This could be used to later
-     * on resume the navigation, for instance as the result of some user response or
-     * equivalent.
+     * To later on resume navigation that was blocked as the result of some user
+     * response, you may save away the new location provided to the blocker function.
```
vicgeralds, line 68: I think what is meant here is "to resume, as the result of some user response" rather than "blocked as the result of some user response". Maybe just remove that?

```suggestion
     * To later on resume navigation that was blocked,
     * you may save away the new location provided to the blocker function.
```

packages/interfaces/src/service/navigation-service.interface.ts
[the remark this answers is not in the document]
vicgeralds: Yeah, you're right about Stencil's watchers. I should know that, but I think I confused them as "before change" watchers, just to make them serve my purpose 😉 😄

packages/interfaces/src/service/navigation-service.interface.ts
vicgeralds: > The problem is rather "who gets to vote"? If we have `blocker A` guarding `content A`, and `blocker B` guarding `content B`, and get a navigation event that will only remove `content B`, then `blocker A` should not get a vote, because `content A` is unaffected by the navigation.

The way I see it, we should then make `blocker A` vote "correctly" rather than somehow prevent it from getting a vote. So then we normally shouldn't write blockers that reject any navigation because they think they "own" the location.

But, I can also imagine a more sophisticated router that would allow specifying each logical route such that you could make it possible to register a blocker to run when navigating away from "your" route.

adrianschmidt: > The way I see it, we should then make `blocker A` vote "correctly" rather than somehow prevent it from getting a vote.

Yeah, we seem to all be in agreement that that's the route to take 👍 🙂

packages/interfaces/src/service/navigation-service.interface.ts
```diff
@@ -56,6 +56,27 @@ export interface Navigator {
      * @return {Location}
      */
     getLocation(): Location;
+
+    /**
+     * Add a blocker that can prevent navigation away from the current page, for
+     * example when there is some unsaved data that would otherwise be discarded.
+     *
+     * If the blocker function returns true navigation is blocked. If it returns
+     * false navigation will proceed as normal.
```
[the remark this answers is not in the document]
civing: Yes, good catch!

packages/interfaces/src/service/navigation-service.interface.ts
```diff
@@ -56,6 +56,27 @@ export interface Navigator {
...
+     *
+     * To later on resume navigation that was blocked as the result of some user
+     * response, you may save away the new location provided to the blocker function.
```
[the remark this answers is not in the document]
civing: Ungh... I 100% agree with you but I do really lack imagination on how to do it. Suggestions? :)

packages/interfaces/src/service/navigation-service.interface.ts
```diff
@@ -85,3 +106,15 @@ export type Location = {
...
+/**
+ * Function to determine if navigation should be blocked or not.
+ *
+ * @param {Location} currentLocation current location before navigation
+ * @param {Location} newLocation location to navigate to
+ * @returns true if navigation should be blocked
+ */
+export type Blocker = (
+    currentLocation: Location,
+    newLocation: Location
```
vicgeralds, line 119: Don't we want them in the other order? First newLocation (to block), then the current location as an additional param?

adrianschmidt: Yes, new value first. That way, if your callback doesn't use the old value, you can just define it with a single argument.

packages/interfaces/src/service/navigation-service.interface.ts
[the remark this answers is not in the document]
vicgeralds:
```javascript
navigator.addBlocker(myBlocker);

let blocked = null;

function myBlocker(newLocation) {
    if (!blocked) {
        showConfirmDialog().then(retry);
    }
    blocked = newLocation;
    return true;
}

function retry(yes) {
    if (yes) {
        navigator.removeBlocker(myBlocker);
        navigator.navigate(blocked);
    }
    blocked = null;
}
```

jgroth (dismissed):

packages/interfaces/src/service/navigation-service.interface.ts
civing: You are so right about this. Pushed a change, please have a look!

packages/interfaces/src/service/navigation-service.interface.ts
```diff
@@ -46,7 +46,7 @@ export interface Navigator {
      * @emits navigate - When the location has been changed the
      * {@link EventDispatcherService} will emit a {@link NavigationEvent}
      *
-     * @param location a new location or parts of the location to replace
+     * @param {Location} location a new location or parts of the location to replace
```
vicgeralds, line 49: This type is wrong and not related to this PR.

vicgeralds (dismissed):

jgroth (approved):

PR #418, opened 2022-04-20
feat(navigation-service): add replace option to navigate method

vicgeralds:
fix Lundalogik/crm-feature#2769


packages/interfaces/src/service/navigation-service.interface.ts
```diff
@@ -80,6 +81,13 @@ export type Location = {
     state: unknown;
 };
 
+export type LocationChange = Partial<Location> & {
+    /**
+     * Specifies whether to replace the current history entry or push a new one
+     */
+    replace?: boolean;
```
jgroth, line 88: This can have three values, `true`, `false`, and `undefined`, but I don't think it's super clear that there is a difference between `false` and `undefined`. How do you feel about something like
```ts
    method?: 'push' | 'replace' | 'auto'
```
Not sure about the names `method` and `auto` though. Perhaps `default` instead of `auto`? I hope you get the idea anyway :slightly_smiling_face:

jgroth: I realize that my suggestion now allows for four values :smile: 
So either
```ts
    method: 'push' | 'replace' | 'auto'
```
which is not great since we don't want to make the setting mandatory, or
```ts
    method?: 'push' | 'replace'
```
and document the behavior of `undefined` better

vicgeralds: I'm starting to think that adding a `push` (and `replace`?) method is clearer still.

jgroth: That would force the consumer to make a decision though, which will make a default behavior hard to implement. Or do you mean in addition to `navigate`?

jgroth (dismissed):

packages/interfaces/src/service/navigation-service.interface.ts
vicgeralds: Yeah, I just thought clear documentation is becoming more difficult. But as long as the default behaviour is desired most of the time, I guess it's not so bad.

jgroth (approved):

PR #421, opened 2022-04-22
feat(navigator): add method for creating URLs for a location

jgroth:
fix Lundalogik/crm-feature#2774


vicgeralds (dismissed):

packages/interfaces/src/service/navigation-service.interface.ts
```diff
@@ -109,6 +109,18 @@ export interface Navigator {
...
+    /**
+     * Create a URL for the given location
+     *
+     * @param {Partial<Location>} location the location to create a URL for.
+     * Any `state` of the location will be ignored.
+     *
+     * @return {URL} the URL to the location
+     */
+    createUrl(
+        location: Pick<Partial<Location>, 'path' | 'query' | 'hash'>
```
adrianschmidt, line 122: I haven't seen `Pick` before, but found the docs for it here: https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys

If I understand correctly, this is basically the same as just copying the `Location` type definition and pasting it here, but removing the `state` property, except that using `Pick`, if we change the type of any of the picked properties on `Location`, that change is automatically applied to this new type too, right?

Is there any more magic to it than that? I can't quite figure out if there's any difference between using `Partial<Location>` and plain `Location`?
- `Pick<Partial<Location>, 'path' | 'query' | 'hash'>`
  vs
- `Pick<Location, 'path' | 'query' | 'hash'>`

jgroth: `Partial` just makes all properties optional, and we don't want to enforce all of them here
I guess we could have used `Omit` as well instead of `Pick`, that might have been more clear since we just want to remove `state`

vicgeralds: Yes, but `Partial` makes all the props optional, so that makes a difference.
The type magic isn't very helpful in the docs...

Here I assumed the TypeScript type and the jsdoc was in agreement.

"Any `state` of the location will be ignored." isn't really the same as "if you try to pass in state it TypeScript may complain"?

jgroth: If I do something like this, it will not complain but ignore `state`
```ts
const location = {
    path: '/',
    query: { foo: 'bar' },
    hash: '123',
    state: { id: 3 }
};
navigator.createUrl(location);
```
It will however complain about the following
```ts
navigator.createUrl({
    path: '/',
    query: { foo: 'bar' },
    hash: '123',
    state: { id: 3 }
});
```

Not sure if you wanted me to change something or not :smile:

vicgeralds: I think it's better to change the type to `Partial<Location>` to keep it simple.

vicgeralds: To avoid type picking, you could also split the location type, like

```typescript
type Location = LocationUrlParts & { state: unknown };
```

and then `createUrl` would instead accept `Partial<LocationUrlParts>` but there's some value in not having too many types with similar names also

vicgeralds: But you decide, I'm OK with merging this :smile:

adrianschmidt: No, I didn't mean to request a change, I was mostly curious about what `Pick` did, and when I learnt that, I wondered why `Partial` would make any difference, but I understand that too now.

But, now I'm wondering about this instead:

> If I do something like this, it will not complain but ignore `state`
> 
> ```ts
> const location = {
>     path: '/',
>     query: { foo: 'bar' },
>     hash: '123',
>     state: { id: 3 }
> };
> navigator.createUrl(location);
> ```
> 
> It will however complain about the following
> 
> ```ts
> navigator.createUrl({
>     path: '/',
>     query: { foo: 'bar' },
>     hash: '123',
>     state: { id: 3 }
> });
> ```

WAT!?

Did you make a typo; am I missing something; or do you really mean that sending in two identical objects will have the compiler complain in only one of the cases, depending on whether or not the object was first stored in a variable!? 🤯

adrianschmidt: > I think it's better to change the type to `Partial<Location>` to keep it simple.

It might depend on the answer to my last comment, but I think I agree with this 😊

jgroth: > Did you make a typo; am I missing something; or do you really mean that sending in two identical objects will have the compiler complain in only one of the cases, depending on whether or not the object was first stored in a variable!? 

No, I don't think I made a typo, and yes this is how it seems to work :smile: 
I don't know how Typescript works internally, but my guess would be that when you use a variable you can use that variable in many places so if the type contains extra stuff those are ignored in the call to the function. However, when you do it inline the data will only be used in that specific place, so any additional data will never be used and is redundant

> I think it's better to change the type to Partial<Location> to keep it simple.

I'm all for making stuff more simple, but I'm not sure if this makes it more complicated or not. I'm fine both with and without `Pick`, but I want to first understand what about it you think makes it more complicated :slightly_smiling_face:

vicgeralds: > WAT!?

It's about excess property checking, which makes sense to avoid mistakes and works for plain objects, but has clear limitations that makes it difficult to rely on.

If it's not a plain object, TypeScript will just handle types as it usually does, if it implements the expected interface it will accept the object that has additional properties.

adrianschmidt: Viktor sent me [this link](https://www.typescriptlang.org/docs/handbook/interfaces.html#excess-property-checks) in Slack, and now I understand.

As long as the consumer can send in a `Location`, and doesn't have to actively remove the `state` property from an existing object, I'm OK with it. But `Partial<Location>` is probably still a lot more helpful to most people reading the docs, and if it does the same thing, isn't it better to opt for simplicity?

jgroth: > But `Partial<Location>` is probably still a lot more helpful to most people reading the docs, and if it does the same thing, isn't it better to opt for simplicity?

That's why I "cheated" and didn't put the full type in the docstring :smile: But I see now that when you read the docs from within VSCode it will still print the whole type. I will change it to `Partial<Location>` to simplify it since you are most likely not going to send in an object literal anyway and the docs state that `state` will be ignored

vicgeralds (approved):

PR #428, opened 2022-04-29
feat(decorators): emit event when `@QueryParam` properties are changed

jgroth:
When a property decorated with `@QueryParam` is changed, an event will be emitted containing the new
value. This will allow clients to automatically reflect the value back to the URL if needed


packages/decorators/src/factory.ts
```diff
@@ -33,6 +34,11 @@ export interface StateDecoratorConfig {
      * @returns {StateOptions} state options
      */
     optionFactory?: OptionFactory;
+
+    /**
+     * Optional watcher to add to the decorated property
+     */
+    watcher?: Watcher;
```
vicgeralds, line 41: Would it be possible to add `watcher` to `StateOptions` instead? Is there a reason to avoid that?

Seems like we have kept `StateDecoratorConfig` minimal before.

packages/decorators/src/state/location.ts
```diff
@@ -1,24 +1,57 @@
...
+    function (value: unknown, oldValue: unknown) {
+        if (isEqual(value, oldValue)) {
+            return;
+        }
+
+        const event = new CustomEvent('changeQueryParam', {
+            bubbles: true,
+            cancelable: true,
+            composed: true,
+            detail: { property: property, value: value },
```
vicgeralds, line 52: Isn't the property something that should be private? If I decorate a private property, I don't expect an event to be emitted with that property name.

I would expect the selector passed with state decorator options to be included with the event instead.

vicgeralds (commented):
I'm thinking, wouldn't it be possible (and nicer) to do this using stencil's `Watch` decorator to implement this? Since we're basically just doing this:

```typescript
@Watch('myQueryParam')
myQueryParamChanged(value, oldValue) {
    if (!isEqual(value, oldValue)) {
        this.changeQueryParam.emit({ key, value })
    }
}
```

If we add the method corresponding to `myQueryParamChanged` to the prototype first (using a unique symbol as property name) and then call `Watch`?

jgroth:
> I'm thinking, wouldn't it be possible (and nicer) to do this using stencil's Watch decorator to implement this? Since we're basically just doing this:

That felt more hackish :smile: The version I showed you last week used this approach. The reason why it feels hackish is because we cannot use the `@Watch` decorator inside here and I had to rely on a lot of Stencil internals in order to get it to work. When a watcher is added to a component, Stencil generates something like this
```ts
class MyComponent {
    static get watchers() {
        return {
            "propA": ["methodA"],
            "propB": ["methodB1", "methodB2"],
        };
    }
}
```

and I had to do something like to following to get it to work

```ts
const watcherName = 'asdqwe123';// generate random method name
const watchers = target.watchers;
watchers[property].push(watcherName);
target[watcherName] = watcher;
```



 It won't build when trying to use the `@Watch` decorator since it seems to not be exported from the Stencil package correctly. Stencil is probably doing something special when using the compiler to build the components

packages/decorators/src/state/location.ts
```diff
@@ -1,24 +1,57 @@
...
  *
  * @experimental
  * @param {StateOptions} options state decorator options
  * @returns {PropertyDecorator} state decorator
  */
 export function QueryParam(options: LocationOptions) {
-    const config: StateDecoratorConfig = {
-        name: PlatformServiceName.Navigator,
-    };
+    return (target: Object, property: string) => {
```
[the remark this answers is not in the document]
jgroth: It's the prototype that will be sent in, not an instance of the class
I'm not sure how to type that correctly, I've tried to Google it but have not found any good answers :smile: 
Do you know how to do it? :slightly_smiling_face:

packages/decorators/src/factory.ts
jgroth: I was not sure where to add it, but `StateOptions` is what goes inside the config to the decorator and is more public. I wanted to hide from any consumers of the decorator. Do you think it makes sense to put it in there?

packages/decorators/src/state/location.ts
jgroth: Yeah, that's true :slightly_smiling_face: You might actually have named your property something else :smile:

packages/decorators/src/state/location.ts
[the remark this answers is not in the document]
vicgeralds: Yeah, that's a bit tricky (I didn't realise it was the prototype first), but in this case `LimeWebComponent` is not a class, it's an interface that prototype implements.

So, as long you don't want the target to be the prototype of a specific class, that's not a problem? I mean where the decorator is used, the type itself doesn't help you using `target` as a prototype rather than an instance.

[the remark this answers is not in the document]
vicgeralds: Anyway, seems like typescript thinks `typeof MyClass.prototype = MyClass` so there's no way to type that you expect an instance of a class, but not its prototype (or the other way around).

Which is crazy when you think about it. You can pass around a prototype and have it mutated as an instance anywhere. 😄

vicgeralds:
> It won't build when trying to use the `@Watch` decorator since it seems to not be exported from the Stencil package correctly. Stencil is probably doing something special when using the compiler to build the components

OK, I understand. If we can't use it directly and need to rely on Stencil internals that's something to try to avoid.

packages/decorators/src/state/location.ts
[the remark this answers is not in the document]
jgroth: 7417ed3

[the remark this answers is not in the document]
jgroth: We should probably change the type of the `target` parameter in the `createStateDecorator` in `factory.ts` as well so you will get an error if you try and use any of the decorators without implementing `LimeWebComponent`. But that's a separate issue :slightly_smiling_face:

packages/decorators/src/state/location.ts
jgroth: e4ac5eb

jgroth:
I decided to rename `property` to `key` in the event 620c783

packages/decorators/src/factory.ts
vicgeralds: I think so, I imagine it would make it a bit cleaner (you wouldn't need to add it to `Property`). It makes sense to keep it less public as well.

Maybe it's that I don't like that we need to create the decorator like `createStateDecorator(options, config)(target, property)` because `config` depends on `options`...

packages/decorators/src/state/location.ts
```diff
@@ -1,24 +1,68 @@
...
  * @returns {PropertyDecorator} state decorator
  */
 export function QueryParam(options: LocationOptions) {
-    const config: StateDecoratorConfig = {
-        name: PlatformServiceName.Navigator,
-    };
+    return (target: LimeWebComponent, property: string) => {
+        const config: StateDecoratorConfig = {
+            name: PlatformServiceName.Navigator,
+            watcher: watchProperty(options.selector),
```
vicgeralds, line 30: Since we're no longer using `property`, we can just return `createStateDecorator(options, config)` directly, right?

vicgeralds (commented):
Great work!

packages/decorators/src/factory.ts
jgroth: I'm a bit confused :smile: 
We agree that we don't want it public, but do you still want to add it to the decorator options? :smile: 

But I think you are right about the syntax as well. I think the code looked a bit different before, forcing me to manually call the factory before doing other stuff. But it seems that can be removed now... I will see what can be done :slightly_smiling_face:

jgroth: 2c3022f

packages/decorators/src/state/location.ts
jgroth: Yes! I didn't like that part of the code either, it's better now :slightly_smiling_face: 
2c3022f

packages/decorators/src/factory.ts
vicgeralds: Yeah, it's like it usually is, we agree what we want, but we can't have it all, so we have to give something up, and that's more difficult 😄

I think it's OK to add it the config. If you think it would be nicer to have it in the decorator options, maybe consider moving it and making it public (in the future)? 🙂

jgroth: Yeah, let's move it and make it public when we realize that we need it :slightly_smiling_face:

packages/decorators/src/watcher.ts
```diff
@@ -0,0 +1,30 @@
...
+    watcher: Watcher
+) => {
+    const descriptor = Object.getOwnPropertyDescriptor(target, property);
+    const originalSetter = descriptor?.set;
+
+    Object.defineProperty(target, property, {
+        set: function (value) {
+            const oldValue = this[property];
+            if (originalSetter) {
+                originalSetter.apply(this, [value]);
```
vicgeralds, line 20:
```suggestion
                originalSetter.callthis, value);
```

packages/decorators/src/watcher.ts
```diff
@@ -0,0 +1,30 @@
...
+            }
+
+            watcher.apply(this, [value, oldValue]);
```
vicgeralds, line 23: here as well (feel free to ignore)

```suggestion
            watcher.call(this, value, oldValue);
```

packages/decorators/src/watcher.ts
jgroth: I don't think I have a preference, just used to `apply` :smile:

jgroth: 7f263d1

jgroth: I think it's because I'm confused by the docs in VSCode, this is what I get from `call`
```
@param argArray — A list of arguments to be passed to the method.
```
It says `array` when it's not supposed to be used as an array :disappointed: 
And `apply` says almost the same :smile: 
```
@param argArray — A set of arguments to be passed to the function.
```

vicgeralds: Hm, yeah that's confusing. 😄

vicgeralds (dismissed):
I think this is good to be merged, with the disclaimer that it probably shouldn't be used where we need to rely on a blocker to prevent changing a query param.

packages/decorators/src/watcher.ts
```diff
@@ -0,0 +1,30 @@
...
+    const originalSetter = descriptor?.set;
+
+    Object.defineProperty(target, property, {
+        set: function (value) {
+            const oldValue = this[property];
+            if (originalSetter) {
+                originalSetter.call(this, value);
+            }
+
+            watcher.call(this, value, oldValue);
```
vicgeralds, line 23: Maybe we could depend on the return value from the watcher to determine if `originalSetter` should be called? It could return the value to set? (but then watcher is not a good name...)

jgroth: How should we determine if a value should not be set? :thinking:

jgroth: I'm not sure I understand :thinking: 
Is it possible to know when we access it from the component and when it's done from the observable?

vicgeralds: Yes, we can save the original setter and use that when creating the state subscription. See fb456aa2969988383882a5edbc1e06da99b68137 for an idea.

We would need something like that if we want to merge both Lundalogik/lime-crm-components#1015 and Lundalogik/lime-crm-components#1022 

I could of course find some better solution than using

```js
@QueryParam({
    selector: '',
    map: [getCardLocation],
});
```
but I still think it should be possible to use `@QueryParam` with `options.map` (it would be even worse if using a non-empty selector string, that would probably cause an infinite loop, passing the mapped state to the watcher, that emits event, that changing the query param to the mapped state, ...)

jgroth: > but I still think it should be possible to use `@QueryParam` with `options.map` (it would be even worse if using a non-empty selector string, that would probably cause an infinite loop, passing the mapped state to the watcher, that emits event, that navigates changing the query param to the mapped state, ...)

Yeah, that should absolutely be possible. Did I break that with this PR? :scream: 
I don't think there will be an infinite loop the way the code is right now. There's both a safeguard in here and I think there is one in the navigator service as well that checks if the values are equal with `isEqual`

vicgeralds: Yes, unless you stop propagating the "changeQueryParam" event (in a parent component), that will change the param to the mapped value, which in theory can be a different value every time. But the main problem is that it can change the query param to the wrong value when we're just reading it. It will also emit "changeQueryParam" the first time when oldValue is undefined

Incidientally, the way I use it in https://github.com/Lundalogik/lime-crm-components/pull/1013/files#diff-bc505e4b049820ab297dc07d80c6e4d28f5e8af2a967f2f33661bda87a7b5959
it is normally the same value, but I still seems to cause an infinite loop, until it is blocked by too many navigations I guess. Seems like some issue with the navigation triggering a synchronous change and the watcher gets called again before the original setter updates the value of the property.

jgroth: Maybe we should just use the `@Watcher` approach anyway :smile: It's an uglier implementation but seems like there is less room for errors as long as Stencil does not break it

PR #434, opened 2022-05-12
Move new decorator

jgroth:


packages/lime-web-components/src/navigator/decorator.ts
```diff
@@ -1,7 +1,4 @@
-import {
-    PlatformServiceName,
-    StateOptions,
-} from '@limetech/lime-web-components-interfaces';
+import { PlatformServiceName, StateOptions } from '../core';
```
adrianschmidt, line 1: I'm guessing this last commit should have been a fixup?

jgroth: Oh yeah, I think I messed up :smile: 
I also noticed that this is more or less the same as #412 :smile:

adrianschmidt (approved):

PR #436, opened 2022-05-13
Refactor testing sub-package

jgroth:
This PR moves the final bits that move code from the old packages to the new one. This moves the `testing` sub-package to new location.

It also contains some other minor fixes, e.g. unintentional breaking changes etc.

Should be rebased on `alpha` once #434 is merged


packages/interfaces/src/limeobject.interface.ts
```diff
@@ -12,5 +12,5 @@ export interface LimeObject extends Base {
      * @param {Property} property the property of the related object
      * @deprecated use `getValue` instead
      */
-    getRelation(property: Property): LimeObject;
+    getRelation?(property: Property): LimeObject;
```
adrianschmidt, line 15: The two changes in this commit seem to fix breaking changes that have already been released previously?

1. Aren't they breaking again then?
2. Shouldn't this commit at least be a `fix`?

packages/decorators/package-lock.json
```diff
@@ -4678,11 +4678,11 @@
         "istanbul-lib-instrument": "^5.1.0",
         "istanbul-lib-report": "^3.0.0",
         "istanbul-lib-source-maps": "^4.0.0",
-        "istanbul-reports": "^3.1.3",
```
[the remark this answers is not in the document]
adrianschmidt: Yeah, this commit disappeared completely now that I updated the lock-files on `alpha` and then rebased 👍

packages/lime-web-components/src/core/platform.ts
```diff
@@ -9,14 +9,14 @@ export interface LimeWebComponentPlatform {
      *
      * @param name the name of the service
      */
-    get(name: PlatformServiceName | string): any;
+    get(name: string): any;
```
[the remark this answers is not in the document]
adrianschmidt: I guess testing this is simply a matter of updating some other packages to use this version, and checking that there are no errors when they try to implement, or fetch the implementation of, these services?

I feel like we can maybe hold off on such testing until we have everything on the `alpha` branch, and then go over and test more or less everything, before we merge `alpha` to `main`?

packages/interfaces/src/limeobject.interface.ts
adrianschmidt: But the bug was introduced [here](https://github.com/Lundalogik/lime-web-components/commit/9f7f598), in v4.6.0, wasn't it? 🤔

jgroth: Hmm :thinking: 
Now I think I'm even more confused, this change should not be needed :smile: 
However, I remember installing everything from `alpha` in one of our repos and got a failing build, and this should have fixed it. It was a few months back though so I'm not quite sure... We could just remove it, but I guess we are going to get a red build when trying the alpha out later :smile:

adrianschmidt: Ah, the other change in this commit at least seems to fix a change that happened earlier on `alpha`:

https://github.com/Lundalogik/lime-web-components/commit/ce8f8fca810e649cda82c6bb152d9959898dbf89#diff-41a1cda5799c5a6abd38ba3442c27c8ecdc1e0eda88617d6d2b454e535a9ad61L36

packages/lime-web-components/src/dialog/dialog.ts
```diff
@@ -42,7 +42,7 @@ export interface DialogRenderer {
     destroy(id: number): void;
 }
 
-export type DialogProperties = Record<string, unknown>;
+export type DialogProperties = Record<string, any>;
```
adrianschmidt, line 45: This one seems to have happened on `alpha`:

https://github.com/Lundalogik/lime-web-components/commit/ce8f8fca810e649cda82c6bb152d9959898dbf89#diff-41a1cda5799c5a6abd38ba3442c27c8ecdc1e0eda88617d6d2b454e535a9ad61L36

packages/interfaces/src/limeobject.interface.ts
adrianschmidt: Summary of Slack conversation: We decided to merge without this change, and will go through and test the `alpha` branch as a whole after this is merged, making any necessary changes for backwards compatibility then.

adrianschmidt (approved):

PR #444, opened 2022-05-25
feat(dialog): add new dialog destroy event for desktop client

jgroth:
When running on desktop and a dialog is opened in a popup window, clients might need to be notified
when the dialog has been closed from an external source such as pressing the X in the corner. A new
event is introduced to solve this need.


packages/interfaces/src/service/dialog-service.interface.ts
```diff
@@ -39,3 +49,9 @@ export interface DialogProperties {
 export interface DialogListeners {
     [listener: string]: (event?: CustomEvent<any>) => void;
 }
+
+/**
+ * @event dialog.destroyed - Emitted when a dialog is externally destroyd, e.g.
```
FredrikWallstrom, line 54: What do we mean by externally destroyed?
This is emitted every time the dialog is destroyed, both on `cancel` button and on the `x` button on the window.

We are manually calling `destroy` when we hit the cancel button in eSign's code, we don't need to do that right because we are doing that in the service now? How does it work in web client's dialog service?

jgroth: > What do we mean with externally destroyed?

When it has been destroyed from the outside, e.g. without anyone calling the `destroy` method. This will happen when the whole popup window is closed.

> This is emitted every time the dialog is destroyed, both on cancel button and on the x button on the window.

Aha! I guess that is because the listener is still there when we do `window.close` and that in turn will invoke the listener that emits the event. I will see if I can fix that

> We are manually calling destroy when we hit the cancel button in eSign's code, we don't need to do that right because we are doing that in the service now? How does it work in web client's dialog service?

Any consumer should still call `destroy` when they want to close the dialog, nothing has changed in that regards to that. This only introduces the event so the consumers can keep track if the dialog is open or not, otherwise it will be hard to keep any open states in sync.

FredrikWallstrom (dismissed):

PR #455, opened 2022-06-14
fix(testing): create empty hook if component's componentWillLoad is undefined

FredrikWallstrom:
Error is thrown if you have multiple tests and the component's `componentWillLoad` is undefined.
This leads to consumers needs to add an empty `componentWillLoad` in their component for the tests to pass.
This will solve this problem.


packages/testing/src/component-testing.ts
```diff
@@ -150,7 +150,7 @@ function extendLifecycleHook(component: any, name: string, hook: () => void) {
 
     const hooks = originalHooks.get(component);
     if (!hooks[name]) {
-        hooks[name] = component[name];
+        hooks[name] = component[name] ? component[name] : function () {};
```
FredrikWallstrom, line 153: To be honest, I'm not into the details about this so not sure if this is the "correct" solution.

jgroth:
I've tried this out but can't get it to work :thinking: 
I still get the `RangeError: Maximum call stack size exceeded` error

FredrikWallstrom:
Hmm ok, I will have to revisit this then when I have some time over.

FredrikWallstrom:
Ok, this works for me 😄 

How did you test this, do you have the test available on Github that I can clone?

FredrikWallstrom:
I scaffolded a new package with `lime-project` and added these tests:

```import { PlatformServiceName } from '@limetech/lime-web-components-interfaces';
import {
    createComponent,
    destroyComponent,
} from '@limetech/lime-web-components-testing';
import { newSpecPage, SpecPage } from '@stencil/core/testing';
import { CoolComponent } from './lwc-limepkg-cool-package-cool-component';

let page: SpecPage;
let component: any;

beforeEach(async () => {
    page = await newSpecPage({ components: [CoolComponent] });
    component = createComponent(CoolComponent, page, {
        tag: 'lwc-limepkg-cool-package-cool-component',
    }) as any;

    const translate = { get: jest.fn() };
    component.platform.register(PlatformServiceName.Translate, translate);

    await page.waitForChanges();
});

afterEach(() => {
    destroyComponent(component);
});

test('component is created', () => {
    expect(component).toBeTruthy();
});

test('component is created', () => {
    expect(component).toBeTruthy();
});
```

jgroth:
Yeah, it seems to work now, I think I screwed up the last time I tested it :smile: 
Great work!

jgroth (approved):

jgroth (approved):

PR #461, opened 2022-06-29
feat(decorators): add a new device type for widescreen

Kiarokh:
Currently we only have mobile, tablet, and desktop,
but that is not enough. We could do more creative things
if we detect larger screen sizes.

fix: https://github.com/Lundalogik/hack-tuesday/issues/247


packages/decorators/src/state/device.ts
```diff
@@ -9,7 +9,7 @@ import { createStateDecorator, StateDecoratorConfig } from '../factory';
  *
  * @type {string}
  */
-export type DeviceType = 'desktop' | 'tablet' | 'phone';
+export type DeviceType = 'widescreen' | 'desktop' | 'tablet' | 'phone';
```
adrianschmidt, line 12: I think "widescreen" works ok, but it might be a bit confusing since it's associated with aspect ratio, not just absolute width. So we might want to try to come up with some other name.

adrianschmidt, line 12: I'm wondering if this might need to be marked as a breaking change…? @jgroth

jgroth: No, it should be ok :slightly_smiling_face: No consumer would get a red build by us introducing this

adrianschmidt: 👌

Kiarokh: Good point. However, since these are for internal use (or?), I think widescreen could be OK. But do you have a better suggestion? `large-screen`? or `large`? :) 

**Why I chose widescreen?**
- widescreen is spelled in 1 word
- indicates that this whole thing is about the screen, not type. I would say the word `deviceType` is what makes this more confusing. Because that connotes that we are detecting the type (like phone, tablet, laptop, desktop, tv, etc…). But this is only about screen width really (right?). I would have preferred names such as: mini > small > medium > large > x-large or similar. But that's too late. However, if we choose `widescreen` here (or something about "screen", like `large-screen` we could break the pattern from "type" to "screen" and make the reader re-think and consider that this is probably not about type detection. (maybe I'm too ambitious here)

Kiarokh: anyhow, suggestions are welcome

adrianschmidt: > I would say the word `deviceType` is what makes this more confusing.

Yeah, ok, I'll give you that.

> * Because that connotes that we are detecting the type (like phone, tablet, laptop, desktop, tv, etc…). But this is only about screen width really (right?). I would have preferred names such as: mini > small > medium > large > x-large or similar. But that's too late.

Yeah, the name `DeviceType` was perhaps not the best, and I think that's on me. But the point here is that those breakpoints are _not_ meant to be static. The idea isn't that `phone` should always be 640px, or whatever it happens to be right now. The idea is that `phone` means "the screen width is about what a phone has nowadays, so you probably want to present things in a way that's good for that kind of size". Or, in other words: "eh, kinda fits-in-a-hand-size-ish?". What number of (css-)pixels that actually is shouldn't be relevant to an add-on, since it shouldn't be using pixels anyway, but relative units. At least that's been my thinking… 😅

Maybe we should simply add `laptop`, and shift the current value for `desktop` to use `laptop` instead, and use a larger breakpoint for `desktop`?

adrianschmidt: I guess it might be breaking once we start using a new value in the decorator implementation though? Since then an add-on could suddenly receive a value it doesn't recognise and doesn't know how to handle…?

Kiarokh: should we go forward with what we have @adrianschmidt ?

adrianschmidt: No, I don't think we should just keep moving forward on a track we don't like because it's difficult to fix the problem. If it's difficult to fix now, it will only be made much harder by adding "widescreen".

I think we should rename `desktop` to `laptop`, and then introduce the new breakpoint as `desktop`. Then we simply mark the change as breaking. Our code should still be responsive enough to handle getting something it doesn't recognise without completely breaking, and if it isn't, we need to address that.

adrianschmidt: Not _definitely_, if proper fallbacks would be in place, but since we have no proper documentation for how to build such proper fallbacks, yes, I agree, we'd have to make it breaking. But I think we should, along with adding some really good docs.

As far as I understand, this PR isn't strictly necessary for the functionality that a user can change the width of the sidePanel? If so, I propose that you make do with what's currently available, and I'll get on with fixing this decorator. Deal? @Kiarokh

adrianschmidt (changes requested):
I'll have a look.

packages/decorators/src/state/device.ts
adrianschmidt: I fully intend to start working on it immediately 😊

jgroth: >So I think we should close this, and if you need the extra breakpoint for the sidePanel before we're done with the new decorator, I would suggest checking the viewport size manually.

Just adding a value here won't solve the issue though, since we still need an implementation that will emit this value at some given breakpoint

adrianschmidt: > > So I think we should close this, and if you need the extra breakpoint for the sidePanel before we're done with the new decorator, I would suggest checking the viewport size manually.
> 
> Just adding a value here won't solve the issue though, since we still need an implementation that will emit this value at some given breakpoint

Absolutely. I'm not sure how that relates to closing this issue though? 🤔

jgroth: > Absolutely. I'm not sure how that relates to closing this issue though? thinking

No, it's not :smile: 
Just wanted to clarify that if we would have merged this we would still have to do more stuff to solve Kias issue :slightly_smiling_face:

adrianschmidt: Ah, ok. Sorry, I didn't mean to "put you in place" or anything, I was just wondering if I was missing something, but didn't express myself very well 😂

PR #467, opened 2022-07-06
feat(testing): expose testing package in a better way

jgroth:
The helpers used for testing should probably not be exposed through the main package after all the re-structuring. It contains both functions that should only be used in tests and a lot of fixture data. Currently, they can only be imported using `@limetech/lime-web-components/dist/testing`, but it would be a bit easier if it was just `@limetech/lime-web-components/testing`

@adrianschmidt I know we have had discussions on this topic in the past :smile: I think this should be fine since it is just for testing and as long as it's documented. Stencil has the same approach, to use their test helpers you have to import them like this
```ts
import { newSpecPage, SpecPage } from '@stencil/core/testing';
```


adrianschmidt:
> I know we have had discussions on this topic in the past 😄 I think this should be fine since it is just for testing and as long as it's documented. Stencil has the same approach, to use their test helpers you have to import them like this
> 
> ```ts
> import { newSpecPage, SpecPage } from '@stencil/core/testing';
> ```

Yeah, I'm fine with that 👍

packages/lime-web-components/package.json
```diff
@@ -12,7 +12,8 @@
     "lime crm"
   ],
   "files": [
-    "dist"
+    "dist",
+    "testing"
```
adrianschmidt, line 16: I've installed all of the subpackages from this branch in lime-crm-components, and I can both build production and run the tests just fine before changing any imports. But when changing the import in a test file (in this case `action-bar.spec.tsx`, but it seems to be the same for all), I get an error while running the tests.

Import changed:
Screenshot showing `import { createComponent, destroyComponent, setState } from '@limetech/lime-web-components/dist/testing';`

Error:
Screenshot showing the rather long error message logged when running the tests, as described in the text above. In short, it complains about `SyntaxError: Unexpected token 'export'`

Is there something I'm missing, or does this require other changes to the consumer?

jgroth: Aha! I never tried to run it since the compiler told me the imports worked :smile: 
This is probably some setup error in crm-components, we might have to look into the jest config if there are some mappings we need to add etc. I will have a look :slightly_smiling_face:

jgroth: Or maybe the build in lime-web-components no longer outputs the same format :thinking: 
I think Jest wants the CommonJS format if I remember correctly, no `export` allowed

jgroth: The `main` and `module` entries were incorrect, it seems the ES5 build is still needed for the tests which is a bit unfortunate, I was hoping to get rid of those... I also found that the entries were incorrect for `package.json` in the new package as well

adrianschmidt: I'm still getting the same error 🙁

jgroth: What does your import look like? If it's `@limetech/lime-web-components/dist/testing` like in your image above that is to be expected since it will use the ES6 version. 
You need to either specify `@limetech/lime-web-components/dist/es5/testing` or the more sane 
`@limetech/lime-web-components/testing` :smile:

adrianschmidt: Well, you never told me that!

# 😄

Nice! I'll update my import and check again 👍

adrianschmidt: Yes, works perfectly 👌

PR #468, opened 2022-07-08
Merge alpha into main

jgroth:


CHANGELOG.md
```diff
@@ -3,6 +3,209 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
```
adrianschmidt, line 5: Hmm… I wonder why the PR checks aren't running for this PR… 🤔

packages/decorators/src/state/application-name.ts
```diff
@@ -1,24 +1,6 @@
...
-    };
-    options.map = [getApplicationName, ...(options.map || [])];
-
-    return createStateDecorator(options, config);
-}
-
-function getApplicationName(applicationData: any) {
-    return applicationData.applicationName;
-}
+export const ApplicationName = SelectApplicationName;
```
adrianschmidt, line 6: For things like this, where we're actually exporting a function, not just an interface, instead of just returning the moved function, I think we could wrap it in a function that logs a deprecation warning, and then returns the result of the moved function… 🤔

```suggestion
export const ApplicationName = (...args) => {
	console.warn('@limetech/lime-web-components-decorators has been deprecated. Please import the ApplicationName decorator from @limetech/lime-web-components instead.')
	return SelectApplicationName(...args);
};
```

Thoughts?

packages/lime-web-components/src/core/plugin-loader.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Interface for the loader component that is loaded when the app starts
+ */
+export interface LimePluginLoader extends LimeWebComponent {
+    /**
+     * This method will be invoked every time the component is inserted into
+     * the DOM. For a plugin loader component, it should happen only once. The
+     * method must be present on the component for technical reasons.
+     */
+    connectedCallback(): void | Promise<any>;
```
vicgeralds, line 12: This should not return a promise. If a promise is returned, stencil will not wait for it.

packages/decorators/src/state/application-name.ts
adrianschmidt: I tried it in the console, and it seems to work just fine:

A screenshot of the dev-tools console, where I have created a simple function with three arguments, then wrapped that in a function like the one suggested in the comment above, along with the result showing that it works.

packages/lime-web-components/src/core/plugin-loader.ts
```diff
@@ -0,0 +1,32 @@
...
+     */
+    componentWillUpdate(): void | Promise<any>;
+
+    /**
+     * This method will be invoked every time the component is removed from
+     * the DOM. For a plugin loader component, it should never happen since
+     * the component will live during the whole lifespan of the application.
+     * The method must be present on the component for technical reasons.
+     */
+    disconnectedCallback(): void | Promise<any>;
```
vicgeralds, line 31: This should not return a promise. If a promise is returned, stencil will not wait for it (and we don't either).

vicgeralds (commented):
I spotted a problem with the new LimePluginLoader interface that we might want to address before merging this.

packages/lime-web-components/src/core/plugin-loader.ts
jgroth: It's not clear what should happen really. I've posted an issue about this in Stencil https://github.com/ionic-team/stencil/issues/3325

packages/lime-web-components/src/core/plugin-loader.ts
jgroth: Same thing here, but I can agree that for `disconnectedCallback` it makes little sense to wait for anything since the element has been removed from the DOM already when this method is invoked

packages/decorators/src/state/application-name.ts
jgroth: Yeah, it should be doable. Do you want to do it here or in a separate PR?

CHANGELOG.md
adrianschmidt: Maybe, but I don't think so. I'm pretty sure I've seen the Snyk test fail and other checks pass. Those should be parallell and not depend on each other as far as I know.

I know I've read about some limit to the number of commits that can be handled somewhere, I think it was for GitHub actions, but I'm fairly sure that was something like 1000, not 149 😄

packages/lime-web-components/src/core/plugin-loader.ts
adrianschmidt: Stencil allows it to return a promise, so it feels to me like we should too… even though it feels a little misleading since it's not actually waited for by anything at the moment…

My guess is that if Johan's issue ever gets resolved, it will probably be that they either fix the problem, so that the promise is actually waited for, or that the update the docs to clarify that it won't be waited for, but still allow the method to return a promise. It's of course possible that they change the method's signature to disallow returning a promise, but that seems less likely to me, since it would likely be breaking for a lot of projects, and just generally be more painful to the Stencil folks 😄

Likewise, even if we haven't had `connectedCallback` in the interface before, people may still have added it to their loader, and if they added one that returns a promise, limiting it to void here would be breaking, wouldn't it?

So, I think the best course of action might be to simply add a note about promises not being waited for, along with a link to the issue perhaps, to the documentation.

Thoughts?

vicgeralds: I think it is clear (that we shouldn't assume something should happen when a promise is returned). In any case, if we say that a promise shouldn't be returned, we remove the source of confusion.

The stencil docs says nowhere that a promise can be returned, but it says that it "follows the same semantics as the one described by the [Custom Elements Spec](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements)". If you read the mdn docs, there's really nothing to suggest that a promise can be returned.

I think there's only confusion about the standard lifecycle callbacks because you mix them up with `componentWillLoad`.

packages/decorators/src/state/application-name.ts
adrianschmidt: I was thinking we should do it before we merge, but I guess that shouldn't really matter, should it?

packages/lime-web-components/src/core/plugin-loader.ts
adrianschmidt: I'm guessing that allowing a promise here is simply a matter of convenience. Without it, you couldn't use `await` without wrapping it in another, non-async, function, I think?

I think this wouldn't be allowed:
```typescript
public async disconnectedCallback() {
    await someFunction();
    continueCleaningStuffUp;
}
```

You'd have to do this instead:
```typescript
public disconnectedCallback() {
	const func = async () => {
          await someFunction();
          continueCleaningStuffUp;
    }
    func();
}
```

If I'm right, that also explains why you might want to allow `connectedCallback` to return a promise, even if you won't wait for it.

CHANGELOG.md
adrianschmidt: Ah! Now I understand! The last commit is called `chore(release): publish v4.50.0-alpha.0 [skip ci]` 🤦‍♂️

When I added `chore: empty commit to try to trigger PR checks`, it ran the checks, but now that I removed it again, the checks again didn't run. Because of the `[skip ci]` obviously 😆

packages/lime-web-components/src/core/plugin-loader.ts
vicgeralds: Actually, just found stencil exports `ComponentInterface`

```typescript
export interface ComponentInterface {
    connectedCallback?(): void;
    disconnectedCallback?(): void;
    ...
}
```

adrianschmidt: [connectedCallback](https://stenciljs.com/docs/component-lifecycle#connectedcallback) is listed under [Component Lifecycle Methods](https://stenciljs.com/docs/component-lifecycle) in the docs, and under [Async Lifecycle Methods](https://stenciljs.com/docs/component-lifecycle#async-lifecycle-methods), it says "Lifecycle methods can also return promises which allows the method to asynchronously retrieve data or perform any async tasks."

That's where the ambiguity comes from.

> But stencil doesn't type check it or provide an interface, right?

Uh, I think I've lived in typescript land for too long… I didn't even consider the possibility that they might not 🤦‍♂️

I'm thinking that limiting the return type here to `void` is a breaking change, if we haven't been limiting it before (since people may already have `connectedCallback`:s in their loaders, even if they weren't part of our interface before. But we maybe also don't want to _suggest_ to people that they should be able to return a promise, as that certainly suggests it could be waited for… so would the best option be to allow `any`? Or maybe just not type the return at all, if that's allowed?

Argh… this is more complicated than I want to handle on a Friday morning! 🤣 Feel free to figure it out between the two of you, I trust your judgement more than I trust my own when it comes to this 😊

adrianschmidt: I think the Stencil docs probably just need this simple change…

```diff
-Lifecycle methods can also return promises which allows the method to asynchronously retrieve data or perform any async tasks.
+Some lifecycle methods can also return promises which allows the method to asynchronously retrieve data or perform any async tasks.
```

adrianschmidt: > Actually, just found stencil exports `ComponentInterface`
> 
> ```ts
> export interface ComponentInterface {
>     connectedCallback?(): void;
>     disconnectedCallback?(): void;
>     ...
> }
> ```

Ah! Nice! I think that settles it then. @jgroth agree?

packages/lime-web-components/src/core/plugin-loader.ts
vicgeralds: I think that's a pattern that should be avoided. If you await something in the disconnected callback, you don't really know that the component is still disconnected, and you're tempted to use `this` and mess with the component.

adrianschmidt: Forget all I wrote after you had found the `ComponentInterface`. I just hadn't seen that comment yet 😅

packages/lime-web-components/src/core/plugin-loader.ts
jgroth: The only reason these are even on the interface to begin with is to solve the annoying "feature" that Stencil removes code from the build that it thinks is not necessary. The whole project that is built needs to have at least one component implementing these methods or the state decorators won't work. (That's only partly true, because the compiler just checks the texts in the files so a comment like `// connectedCallback` would make the compiler include the necessary code :smile: )
With this in mind, we can also just be vague and remove the return type altogether :smile:

vicgeralds: > ```diff
> -Lifecycle methods can also return promises
> ```

I missed that! I agree that creates confusion.

adrianschmidt: > With this in mind, we can also just be vague and remove the return type altogether 😄

If that's an option, I honestly think that sounds like a good way forward.

packages/decorators/src/state/application-name.ts
adrianschmidt: It's not that many, so I can just as well do it now, and we can get it tested before merging.

jgroth: Great! :slightly_smiling_face:

jgroth:
Can we merge this? :smile:

adrianschmidt:
Seems it already doesn't have that rule active. Unless you just changed it.

PR #471, opened 2022-07-11
fix(decorators): emit context changes correctly

vicgeralds:
fix: Lundalogik/crm-feature#2891


packages/lime-web-components/src/core/decorators/factory.spec.ts
```diff
@@ -135,15 +126,12 @@ describe('createStateDecorator', () => {
...
-                expect(true).toBeTruthy();
-            });
-            component.platform = platform;
-            component.context = context;
+        it('returns a promise that resolves when platform and context has been set', async () => {
+            const result = component.connectedCallback();
             element.platform = platform;
             element.context = context;
+
+            return result;
```
jgroth, line 134: Is this a refactor or is there a new behavior with the new implementation?
I don't think the test needs to be `async` since the result should be a `Promise` anyway. Maybe we should add
```ts
expect(result).toBeInstanceOf(Promise)
```
just to be sure?

vicgeralds: Yes, this is a refactor, as I think I didn't really have to change this particular test. 

I'll try to revert such changes from this commit.

jgroth (dismissed):

PR #472, opened 2022-07-11
fix(loader): remove return types

jgroth:
The Stencil docs are a bit unclear about return types for some of the lifecycle hooks. These hooks
are only included on this interface to make sure they are included in each project, so that Stencil
does not remove the necessary code from the build. By removing the return types we make the
interface more vague but also make it future proof.


packages/lime-web-components/src/core/plugin-loader.ts
```diff
@@ -9,24 +9,24 @@ export interface LimePluginLoader extends LimeWebComponent {
...
      */
-    componentWillLoad(): void | Promise<any>;
+    componentWillLoad();
 
     /**
      * This method will never be called since the component will never
      * be updated. It must still be present on the component though.
      */
-    componentWillUpdate(): void | Promise<any>;
+    componentWillUpdate();
```
vicgeralds, line 23: I expect that this can be removed after fixing Lundalogik/crm-feature#2891

vicgeralds (approved):
Looks good

packages/lime-web-components/src/core/plugin-loader.ts
jgroth: Yeah, if we are not using it in the decorators it can probably be removed. It's no harm in leaving it either :slightly_smiling_face:

PR #478, opened 2022-07-12
fix: include Stencil in build

jgroth:
fix Lundalogik/crm-feature#2898

`getElement` from Stencil is used by the decorators and is needed in the build


packages/lime-web-components/package.json
```diff
@@ -40,5 +40,11 @@
     "@limetech/lime-web-components-decorators": "^4.51.1",
     "@limetech/lime-web-components-interfaces": "^4.51.1",
     "@limetech/lime-web-components-testing": "^4.51.1"
+  },
+  "peerDependencies": {
+    "@stencil/core": "^2.17.1",
+    "@types/jest": "^27.5.0",
+    "jest": "^27.5.1",
+    "jest-cli": "^27.5.1"
```
vicgeralds, line 48: To me it doesn't make any sense to have jest as a peer dependency. Why?

vicgeralds (changes requested):
Good to have stencil as a peer dependency. 👍
Not sure why we need this when we didn't before 😄 

The package-lock files are not in sync (means `npm ci` fails) and seem generated by lerna (reformatted with different indentation than when you do `npm install`).

packages/lime-web-components/package.json
Befkadu1: What is the problem having jest in `devDependencies`?

vicgeralds: @Befkadu1 no one said there was a problem having jest in devDependencies. 🙁

vicgeralds: He moved them from `peerDependencies`, I can see the comment is confusing without the original context

vicgeralds (dismissed):
I updated the package-lock.json files 🙂

jgroth:
> I updated the package-lock.json files slightly_smiling_face

Great! :slightly_smiling_face:

PR #481, opened 2022-07-14
fix(build): inline ts helpers for ES5

jgroth:
Not sure how or why, but the tests in Lundalogik/lime-crm-components#1124 are failing when some decorators are used. The config for the old `decorators/` build looks the same as the build in the new package, and using the old package works. Inlining all the TS helpers for ES5 seems to fix the issue


vicgeralds:
Strange indeed. I'm trying to understand what's going on,  "using the old package works" for me as well, even when I can see that `importHelpers` is `true` and I see references to `tslib_1.__spreadArray` in the build...

jgroth:
I tested your suggestion and it seems to work :slightly_smiling_face:

jgroth:
I added it to `devDependencies` now without thinking about it. Maybe it needs to be under `dependencies`? :thinking:

vicgeralds (approved):

PR #498, opened 2022-08-08
fix(testing): inject component states in connectedCallback

civing:
Instead of injecting states for unit testing in componentWillLoad inject
them in connectedCallback lifecycle method.

Some unit tests need the injected state already in connectedCallback.
Since they were not injected already some components in the production
code solves testability by having componentWillLoad call
connectedCallback. It's not a good practice to call a lifecycle method
inside another one. After this commit it should not be necesarry to do
so


packages/lime-web-components/src/testing/component-testing.ts
```diff
@@ -62,7 +62,7 @@ export function createComponent(
     componentConfigs.set(component.prototype, config);
     extendLifecycleHook(
         component.prototype,
-        'componentWillLoad',
+        'connectedCallback',
```
civing, line 65: I'm not 100% sure this lifecycle method is well suited for state injection since it will run every time the component is reattached to the DOM. Will this even happen outside some very elaborate unit test? If so, could we add some hidden state to the component to keep track if the states has been injected or not? ⁉️

On the other hand, forcing components to call a lifecycle method inside another lifecycle method (componentWillLoad -> connectedCallback) is arguably worse in my humble opinion.

adrianschmidt: I don't think we need to worry too much about the performance of disconnecting and then reattaching components in tests. If that's what you're worried about?

> On the other hand, forcing components to call a lifecycle method inside another lifecycle method (componentWillLoad -> connectedCallback) is arguably worse in my humble opinion.

Yes, that definitely sounds worse. Especially if that's something we're doing generally, not just in tests?

civing: I'm worried about that `initComponent` will be run for every `connectedCallback`. Some intricate test case might be screwed if we would run `initComponent` after every attach.

> Yes, that definitely sounds worse. Especially if that's something we're doing generally, not just in tests?

If you search for connectedCallback in for instance `lime-crm-components` you will find loads of these:
```javascript
public componentWillLoad() {
  this.connectedCallback();
}
```

adrianschmidt: Thinking about it, I don't think that's likely to actually be a problem, but I would probably prefer something like the following, which I'm fairly sure we also use in places:

```typescript
public componentWillLoad() {
	this.initialize();
}

public connectedCallback() {
	this.initialize();
}

private initialize = () => {
	…do stuff…
}
```

adrianschmidt: Come to think of it, I think `componentWillLoad` is the correct place to inject component state, because when a component first loads, `connectedCallback` is called first, and at that time, the state has not been injected, which is the whole reason we often need to run setup in both `connectedCallback` _and_ `componentWillLoad`.

So by injecting the state in `connectedCallback`, we open up the vast majority of our tests to false negatives; components that do their initialisation only in `connectedCallback`, but would actually have to run it in `componentWillLoad` too, will work just fine in test, but will fail in reality.

adrianschmidt: > But it happens, for instance when we use some decorators. In those cases the states will be fetched prior to `connectedCallback` but in the tests it would be injected before(?) `componentWillLoad` One example from the code were we need the state already in the connectedCallback. The state is constructed with the config decorator in production.
> 
> ```js
>    @State()
>     @Configs({
>         name: 'webclient_keyboard_shortcuts',
>     })
>     private config: KeyboardShortcutsConfig;
> 
>     public connectedCallback() {
>         if (!this.config) {
>             return;
>         }
> 
>         if (!this.subscription && !this.isEmptyContext(this.context)) {
>             this.subscription = this.focusedContextObserver
>                 .getObservable()
>                 .subscribe(this.handleContextFocusChange);
>         }
> 
>         this.updateKeybinding();
>     }
> 
>     public componentWillLoad() {
>         this.connectedCallback();
>     }
> ```

Have you actually checked that `this.config` is set the first time `connectedCallback` is called? (I haven't myself, so no shade if you didn't.) I'm thinking the `if (!this.config) { return; }` and the calling of `connectedCallback` from `componentWillLoad` is a fairly strong indicator that `this.config` isn't set when `connectedCallback` is called the first time… 🤔

The PR description says "Some unit tests need the injected state already in connectedCallback." I'm a little doubtful that's true. There are no tests failing currently, so I assume the tests mentioned are skipped, on a branch somewhere, or hypothetical? If they exist, are we sure they are even correct?

civing: Yeah, we found this issue when adding unit tests to one of @Befkadu1's PRs. Config was set in the browser but due to the lack of this PR not set in the tests. But now that you question it I start to second doubt that 😅

My theory is that calling `connectedCallback` from `componentWillLoad` was added to make the code testable.

adrianschmidt: > My theory is that calling `connectedCallback` from `componentWillLoad` was added to make the code testable.

Hmm… I hope that isn't the case… structuring code so that it's testable is one thing, but changing the behaviour of code just to make it testable, because the test doesn't represent reality, isn't ideal… 😞

I think I've also had cases where some "state" was accessible already in the initial `connectedCallback`, so it's very possible that that _is_ the case when using decorators. It's not the case when using `@Prop` though, unless something changed very recently. So I don't think it's a good idea to change so that we inject values for `@Prop`s in the first `connectedCallback`. I guess it's possible that we should do so in any subsequent `connectedCallbacks`… but now I'm starting to feel like we're mocking too much of what the browser is doing.

On that topic: Is the test in question a `spec` or `e2e` test? If it's a `spec`, could we avoid having to manually inject stuff like this by converting it to an `e2e` test instead, so it's run in an actual browser?

civing: Yeah, I'm a bit unsure however my fix is too much of a hack. At least it fixes the test framework in such a way that we can remove the workarounds in the production code. That is calling a lifecycle method from another lifecycle method.

> On that topic: Is the test in question a `spec` or `e2e` test? If it's a `spec`, could we avoid having to manually inject stuff like this by converting it to an `e2e` test instead, so it's run in an actual browser? (Maybe that's not at all the problem here—I'm just brainstorming a little 😅)

I don't think that would make a difference in this case though. 😞
In that case we need the decorators to inject the states and I don't think running in E2E mode would make them do that. If I'm not missing something fundamental that is 😊

civing: Maybe @jgroth can give us a second opinion here?

adrianschmidt: > At least it fixes the test framework in such a way that we can remove the workarounds in the production code. That is calling a lifecycle method from another lifecycle method.

If I recall correctly from two weeks ago, my whole argument is that this _isn't_ fixing the test framework. It's making the test framework do something that isn't happening in reality, so it's actually _breaking_ the test framework. I can't speak for the specific case you're looking at, but in general, having `componentWillLoad` call `connectedCallback`, or having both of them call a third function `initialize`, is not done because it's required to make the code testable, it's done because it's necessary for the code to work properly when run _for real_ in the browser, by a user, in production.

civing: Yeah okey. 👍 
When me and @Befkadu1 tested around it worked in the browser but we needed to call `connectedCallback` from within `componentWillLoad` to make it testable.

adrianschmidt: Yeah, it depends on what the component is doing while setting up. If it does anything that needs to be cleaned up on teardown, like adding event listeners manually (Stencil makes sure stuff like `@Watch` is cleaned up automatically), then you need to do the teardown in `disconnectedCallback`. But we can never know if `disconnectedCallback` is called because the component is being destroyed (_almost_ always the case) or because it's just being disconnected from the DOM in order to be connected again later, or in a different place (Stencil doesn't do this, but we can't know that someone doesn't write code that moves components around, so we have to take it into account). The problem is that if the component is reconnected again, only the `connectedCallback` will fire, not the `componentWillLoad`.

So, in your case (looking at the value of decorators, if I recall correctly?), the real world case only required you to do something in `connectedCallback`, because the values you needed have already been loaded in the very first `connectedCallback`.

The problem is that when you are looking at the value of `@Prop`s, the values are not available in the first `connectedCallback`. So you have to run the setup from `componentWillLoad`. But `componentWillLoad` won't run if the component is disconnected and then reconnected. So you have to run the setup from `connectedCallback` too. But if the test framework injects _all_ values in `connectedCallback`, a component reading from a `@Prop` will pass the tests but fail in reality.

Optimally, I guess we should make it possible to supply mock values in different groupings, one for decorator values, and one for props (and any others too, if there are other kinds(?)), and make them available in the phase they would be available in in reality.

Maybe, with the work you've done here, it wouldn't be too difficult to make that happen?

jgroth: :dizzy_face: 

There's a lot to read here, but I will try and give my 2¢ before doing that :smile: 

It's hard to write simple tests at the moment. There's too much boilerplate and you can quite often run into use cases where you have to create a lot of mocks and you end up copy/pasting a lot from other components. I want to make this easier. Part of that is probably to remove these hackish test functions altogether and do something more simple. 

The idea is to add a new platform type in limeclient.js, e.g. `TestPlatform`, that uses all the real service implementations and uses [mswjs](https://mswjs.io/) to mock network calls. This platform can be used when writing tests so we don't have to mock so much. It can also be used for creating live examples in documentation, i.e. like we have in lime-elements which probably would add a lot of value for others using the platform.

With this in mind, I _think_ all you would have to do in a test is to write something like this to setup your component

```tsx
import { newSpecPage } from '@stencil/core/testing';
import { createTestPlatform } from '@lundalogik/limeclient.js/testing';
import { MyCmp } from '../my-cmp';

let component;
let platform;
let context;

beforeEach(async () => {
    platform = await createTestPlatform();
    context = { limetype: 'deal', id: 1001 };
    const page = await newSpecPage({
        components: [MyCmp],
        template: <my-cmp platform={platform} context={context} someProp="test"/>,
    });
    await page.waitForChanges();
    component = page.root;
});
```

Note that there is no magic setup procedure here. All of it is using standard Stencil stuff and the component will be initialized with all the values given in the template.

jgroth: > My theory is that calling `connectedCallback` from `componentWillLoad` was added to make the code testable.

I very much hope this is not true :scream: 
If you have to modify your source code just to get the same behavior in the tests as running the code in the browser then there's obviously something wrong and a bug that needs to be fixed :smile: 

Have you tested this change in crm-components? Do all the tests pass?

civing: I can't remember now if I tested this change on all tests or only @Befkadu1 change...So I don't know if this change will brake these tests calling lifecycle methods from within a lifecycle method. However, when we have this change we can remove that behaviour from the production code alltogether 😄

civing:
Merging this PR will fail two tests in lime-crm-component. Those will be easily fixed by removing the chained life cycles in the production code. Here is a PR that does that https://github.com/Lundalogik/lime-crm-components/compare/fix_calling_of_lifecycle_hooks?expand=1

```
Summary of all failing tests
 FAIL  src/dock/dock.spec.tsx (6.766 s)
  ● registers the correct keybind

    expect(jest.fn()).toHaveBeenCalledTimes(expected)

    Expected number of calls: 1
    Received number of calls: 2

      105 |
      106 | test('registers the correct keybind', () => {
    > 107 |     expect(keybinds.bind).toHaveBeenCalledTimes(1);
          |                           ^
      108 |     expect(keybinds.bind).toHaveBeenCalledWith('Ctrl+O', command);
      109 | });
      110 |

      at Object.<anonymous> (src/dock/dock.spec.tsx:107:27)

 FAIL  src/action/action-button/action-button.spec.tsx (7.288 s)
  ● it registers the correct keybind

    expect(jest.fn()).toHaveBeenCalledTimes(expected)

    Expected number of calls: 1
    Received number of calls: 2

      81 |
      82 | test('it registers the correct keybind', () => {
    > 83 |     expect(keybinds.bind).toHaveBeenCalledTimes(1);
         |                           ^
      84 |     expect(keybinds.bind).toHaveBeenCalledWith('ctrl+alt+delete', command);
      85 | });
      86 |

      at Object.<anonymous> (src/action/action-button/action-button.spec.tsx:83:27)


Test Suites: 2 failed, 160 passed, 162 total
Tests:       2 failed, 9 skipped, 1366 passed, 1377 total
Snapshots:   0 total
Time:        24.249 s
Ran all test suites.
```

packages/lime-web-components/src/testing/component-testing.ts
adrianschmidt: I still don't think you've understood what I'm worried about here. I think we need to have a call. I'll ping you both in Slack.

adrianschmidt: I was the one not understanding 😁

Resolving!

adrianschmidt (approved):

PR #531, opened 2022-09-15
Missing types

jgroth:
This PR contains a few missing pieces I have noticed that would be good to get in before we make a breaking change.

1. The first commit contains an interface describing errors that are thrown by the `HttpClient`. The error has been thrown all the time by the web client, but it has been hard to know how the error looks.
2. Commands need to be created from configs a lot of the time and to do that there must be some way to get the class tied to the command in order to create it. The web client implementation has had a `getCommandClass` method on the service that has been used (typing the service to `any`), but a better solution is probably to just add a factory method to the command bus that will create the commands from the config right away.
3. The web client has some internal commands used for navigating the app. Those have recently been duplicated in crm-components, and they are generic enough to be useful for a lot of consumers. It will be a lot easier to configure actions with an already existing command, instead of each consumer having to create their own implementation. 

Numbers 2 and 3 of these are "needed" to fix Lundalogik/lime-crm-components#1234


Kiarokh:
Great work @jgroth! 
Shouldn't we put this as a separate PR in the review column, on the board? 

I'm worried that no one dares to review the other PR :)

PR #545, opened 2022-11-03
Use tsdoc and API extractor

jgroth:
This PR replaces jsdoc with tsdoc and updates all doc blocks accordingly. It also uses [API extractor](https://api-extractor.com/) to verify the public API when building on GitHub.


jgroth:
@adrianschmidt feel free to add yourself to the `.github/CODEOWNERS` file if you want. The way it has been set up now I _think_ I would be requested for a review if there are any API changes :smile:

package.json
```diff
@@ -43,13 +43,14 @@
...
     "eslint-plugin-ban": "^1.6.0",
-    "eslint-plugin-jsdoc": "^37.9.7",
     "eslint-plugin-prefer-arrow": "^1.2.3",
     "eslint-plugin-prettier": "^4.2.1",
     "eslint-plugin-react": "^7.31.8",
     "eslint-plugin-sonarjs": "^0.15.0",
+    "eslint-plugin-tsdoc": "^0.2.17",
     "jest": "^27.5.1",
     "jest-cli": "^27.5.1",
+    "jju": "^1.4.0",
```
adrianschmidt: 🤔

This doesn't seem to be used anywhere (at least not in this commit, I'm reviewing on a per-commit basis). Does `eslint-plugin-tsdoc` have a peer dependency requirement on it?

jgroth: I thought I removed that :thinking: 
I think there was a bug in an earlier version because it complained about this library missing so I added it to get it to work. Yesterday when I updated to the latest version I found that I could remove this dependency, so not sure how it ended up here again :smile:

src/application/decorators/session.ts
```diff
@@ -8,8 +8,9 @@ import { createStateDecorator } from '../../core/decorators/factory';
 /**
  * Get the application session
  *
- * @param {StateOptions} [options] state decorator options
- * @returns {Function} state decorator
+ * @param options - state decorator options
+ * @returns state decorator
```
adrianschmidt, line 12: I was like "Wait, no indication that `options` is optional? And ny type information? Eh, wait-a-minute! _'tsdoc'_… 🤦‍♂️🤦‍♂️"

# 🤣

package.json
jgroth: Oh it is removed. I probably screwed up the rebase because it's been removed in another commit in this PR

src/userdata/repository.ts
```diff
@@ -3,30 +3,30 @@ import { StateRepository } from '../core';
...
      */
     get<T = any>(key: string): T;
 
     /**
      * Set user data
      *
-     * @param key name of the key for the data
-     * @param data the data to save, will delete the data if undefined
+     * @param key - name of the key for the data
+     * @param data - the data to save, will delete the data if undefined
```
adrianschmidt, line 29: Waaaa… this was a _lot_ of menial changes… did you do all these by hand (😵), or did you automate it in some way?

src/application/decorators/session.ts
jgroth: :smile: 

Yeah it feels a bit unnecessary to duplicate all that in the doc block when we can get that information from the code anyway :slightly_smiling_face:

package.json
```diff
@@ -50,14 +51,16 @@
     "eslint-plugin-tsdoc": "^0.2.17",
     "jest": "^27.5.1",
     "jest-cli": "^27.5.1",
-    "jju": "^1.4.0",
```
adrianschmidt: Ah 😄

This specific line should be squashed into the earlier commit instead. If that's less than trivial for you to do, I can do it, or show you how to do it (fairly) easily 🙂

adrianschmidt: (Of course, I'm going to be a pita and expect the package-lock.json to be in sync too… sorry! 🙈)

package.json
```diff
@@ -50,14 +51,16 @@
...
     "prettier": "^2.7.1",
     "replace-in-file": "^6.3.5",
     "shelljs": "0.8.5",
     "typedoc": "0.17.7",
     "typedoc-plugin-markdown": "~2.4.2",
     "typescript": "^4.8.3"
   },
+  "peerDependencies": {
+    "@stencil/core": "^2.18.0"
```
adrianschmidt, line 64: Ooh… this will tell our consumers they need to use Stencil 2.18.0 or higher in their projects… that's going to mean a lot of updating for a lot of add-ons…

.github/CODEOWNERS
```diff
@@ -0,0 +1,3 @@
+.github/CODEOWNERS  @jgroth
+
+etc/*               @jgroth
```
adrianschmidt, line 3: I'll move this to its own commit and add myself to it as well.

src/userdata/repository.ts
adrianschmidt: Good to hear! 😅

package.json
jgroth: Ok, will try to fix it :slightly_smiling_face:

package.json
jgroth: This is probably another rebase issue, I don't think I explicitly added this for this PR :smile:

package.json
jgroth: I think I fixed it. Did not manage to do it as a fixup though because it would end up with an empty commit. Maybe there is a trick to make that work? :thinking:

adrianschmidt: No worries! 👍

package.json
jgroth: Removed it in a fixup. It's listed as a dependency anyway so no reason to have both, not sure how it ended up here

adrianschmidt (approved):
Seems to work just fine to me 👍

PR #549, opened 2022-11-23
feat(limeobject): add validation errors to exception interface

jgroth:
fix Lundalogik/crm-feature#2956


src/limeobject/error.ts
```diff
@@ -5,5 +5,23 @@ import { LimeObject } from './limeobject';
...
+}
+
+/**
+ * @public
+ */
+export interface ValidationError {
+    /**
+     * Name of the field the error belongs to
+     */
+    [key: string]: string[] | ValidationError;
```
adrianschmidt, line 26: So, I suppose this is either `nameOfSection: <ValidationError>` or `nameOfField: [one or more strings forming the error message]`?

It's not super clear from the documentation provided.

adrianschmidt, line 26: Also, for some reason, the docblock here doesn't get included in the docs 🤔

>     /**
>      * Name of the field the error belongs to
>      */

image

jgroth: Yeah, not sure how to get that in there :thinking: 
I can maybe add an example on the interface to make it a bit more clear

adrianschmidt: The updated docs look great!

adrianschmidt: This line still isn't included in the generated documentation for whatever reason (perhaps because the property uses `[key: string]`?), but with the description and example you added for the interface itself, I don't think that's a problem 👍

adrianschmidt (approved):

PR #588, opened 2023-01-17
feat(userpreferences): add user preferences interface

Befkadu1:
fix: https://github.com/Lundalogik/crm-feature/issues/3139


src/userpreferences/repository.ts
```diff
@@ -0,0 +1,23 @@
...
+     */
+    get<T = any>(key?: string): T;
+
+    /**
+     * Set user data
+     *
+     * @param key - name of the key for the data
+     * @param data - the data to save, will delete the data if undefined
+     */
+    set<T = any>(key: string, data?: T): void;
```
civing, line 16: Shouldn't this return a promise?
```suggestion
    set<T = any>(key: string, data?: T): Promise<void>;
```

adrianschmidt: I guess that depends on whether we want to be able to tell the consumer that something went wrong. If we consider it a fire-and-forget type action, we don't need a promise. But, since certain changes may require the user to either reload the app, or log out and log back in again, we probably want to know when the save is done, so the page can safely be reloaded…

The consumer can always just ignore the promise if they don't care, so it seems like a good idea to return at promise 👍

src/userpreferences/repository.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * Handle fetching and saving user preferences data
+ * @alpha
+ */
+export interface UserPreferencesDataRepository {
```
adrianschmidt, line 5:
```suggestion
export interface UserPreferencesRepository {
```

src/userpreferences/repository.ts
```diff
@@ -0,0 +1,27 @@
...
+     */
+    get<T = any>(key?: string): T;
+
+    /**
+     * Set user data
+     *
+     * @param key - name of the key for the data
+     * @param data - the data to save, will delete the data if undefined
+     */
+    set<T = any>(key: string, data?: T): void;
```
adrianschmidt, line 20: I'm thinking, should we really support generic types here? Sure, it makes it a tiny bit easier for the consumer, but is it really worth the extra complexity it puts on us for that very minor benefit? We could just type them to `string`, and have the consumer do any conversion to and from json. If _we_ do the conversion, we also need to handle any cases where the data might not be possible to stringify or parse…

src/userpreferences/types.ts
```diff
@@ -0,0 +1,21 @@
...
+import { UserPreferencesDataRepository } from './repository';
+
+const SERVICE_NAME = 'state.user-preferences';
+
+PlatformServiceName.UserPreferencesDataRepository = SERVICE_NAME;
+
+declare module '../core/platform' {
+    interface PlatformServiceNameType {
+        /**
+         * @see {@link UserPreferencesDataRepository}
```
adrianschmidt, line 11:
```suggestion
         * @see {@link UserPreferencesRepository}
```

src/userpreferences/types.ts
```diff
@@ -0,0 +1,21 @@
...
+         */
+        UserPreferencesDataRepository: typeof SERVICE_NAME;
```
adrianschmidt, line 13:
```suggestion
        UserPreferencesRepository: typeof SERVICE_NAME;
```

src/userpreferences/types.ts
```diff
@@ -0,0 +1,21 @@
...
+    }
+
+    interface LimeWebComponentPlatform {
+        get(
+            name: PlatformServiceNameType['UserPreferencesDataRepository']
```
adrianschmidt, line 18:
```suggestion
            name: PlatformServiceNameType['UserPreferencesRepository']
```

src/userpreferences/types.ts
```diff
@@ -0,0 +1,21 @@
...
+        ): UserPreferencesDataRepository;
```
adrianschmidt, line 19:
```suggestion
        ): UserPreferencesRepository;
```

src/userpreferences/repository.ts
Befkadu1: I agree!
@adrianschmidt are you suggesting `set<T = any>(key: string, data?: Record<string, string>): Promise<void>;`?

adrianschmidt: Ah, no, I suggest making it simpler instead:

```diff
-    set<T = any>(key: string, data?: T): Promise<void>;
+    set(key: string, data: string): Promise<void>;
```

etc/lime-web-components.api.md
```diff
@@ -793,6 +793,13 @@ export interface UserGroup {
     name: string;
 }
 
+// @alpha
+export interface UserPreferencesDataRepository {
```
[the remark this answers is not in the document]
adrianschmidt: You need to run `npm run build` locally for this file to be updated.

src/userpreferences/repository.ts
adrianschmidt: Fixed by Bef and verified by me, so I'm resolving this thread 🙂

src/userpreferences/repository.ts
Befkadu1: I think you misunderstood it. The `key` is the column name in `sys-data` table. We can't save for example the string `sv`. We suppose to save like this `language: sv`

adrianschmidt: Aha… but, shouldn't the `sys_data` key be hardcoded? Do we ever need to change that? 🤔

I thought we decided to store all user preferences in a single big json object in `sys_data`, under a single key?

Befkadu1: Yea you are right! I will take your suggestion. @civing FYI I am going to make a change here. We need to make changes in the other PRs in limeclient.js as well. I will do that now so that you don't get surprised by the changes😆

adrianschmidt: I'm just reviewing the PR in limeclient.js now, and that was already built with a hard coded key. So I think the only change needed there should be to change the generic type to `string` 👍

etc/lime-web-components.api.md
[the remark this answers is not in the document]
civing: I'll have to look at it, I just ran rpm run api:update yesterday so I have no idea why it should be wrong 🤪

src/userpreferences/repository.ts
```diff
@@ -15,9 +15,9 @@ export interface UserPreferencesRepository {
      * Set user data
      *
      * @param key - name of the key for the data
-     * @param data - the data to save, will delete the data if undefined
```
adrianschmidt: I think we can let this stay in. Looks like `JSON.stringify()` removed properties with `undefined` as the value:

image

src/userpreferences/repository.ts
civing: hm, okey! This might be a fair point. The assumption was that the user of this API would like to store an object under a certain key, but they could just serialize it themselves I guess.

etc/lime-web-components.api.md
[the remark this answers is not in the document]
adrianschmidt: There have been changes made since you ran that 😆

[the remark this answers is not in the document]
adrianschmidt: Ah, does `npm run build` not also run the api update stuff? I didn't know that. Then scratch my comment stating that you need to build locally, what you need to do is run `npm run api:update` 👍

src/userpreferences/repository.ts
jgroth: > I'm thinking, should we really support generic types here? Sure, it makes it a tiny bit easier for the consumer, but is it really worth the extra complexity it puts on us for that very minor benefit?

> So I think the only change needed there should be to change the generic type to string

I don't think I agree. What extra complexity do you see here that would make it hard to have a generic? :thinking: We would just have to return whatever is saved under a specific key, or am I missing something? If we want to make sure it is serializable we can do something like

```ts
type JsonData =
    | string
    | number
    | boolean
    | null
    | undefined
    | JsonData[]
    | { [key: string]: JsonData };

interface Foo {
    set<T extends JsonData = any>(key: string, data?: T): void;
}
```

etc/lime-web-components.api.md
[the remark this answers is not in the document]
Befkadu1: This one `npm run api:update` works 😆

src/userpreferences/repository.ts
adrianschmidt: If we require it to be a string for now, will it be a problem if we decide to allow other types later? @jgroth

src/userpreferences/repository.ts
```diff
@@ -0,0 +1,27 @@
...
+ * @alpha
+ */
+export interface UserPreferencesRepository {
+    /**
+     * Get user preferences
+     *
+     * @param key - name of the key containing the data
+     * @returns the data
+     */
+    get<T = any>(key?: string): T;
```
adrianschmidt, line 12: Leave this be until the discussion with Johan about generic type has come to some conclusion, but: _if_ we remove the generic type from set, it should also be removed from here:

```suggestion
    get(key?: string): string;
```

Befkadu1: Yes let us conclude the discussion with Johan first

src/userpreferences/repository.ts
jgroth: Probably not, I just think it's overly restricted to only allow strings when the implementation for returning any other value will be the same :smile: 
I'm also still not sure about the complexities you mentioned :slightly_smiling_face: 

If it's the generic you don't like we could also do it like this

```ts
interface Foo {
    set(key: string, data?: JsonData): void;
}
```

adrianschmidt: Yeah, it's partly the requirement for the consumer to set a generic type, especially when getting the value, and partly that I felt like we might need to handle cases where data is sent in that can't be serialised… maybe I'm overthinking that part, but isn't there data structures in javascript that can use objects as keys and stuff, nowadays? Can those really be serialised into json?

On the other hand, I don't want to drag out a long discussion about it, I just want the interface to be small and safe, and not give us headaches in the future 😄

adrianschmidt: 😬

image

jgroth: With the suggested type above your example would give a compile error, but it would still be possible to add objects/arrays/primitives. I think that would be just as safe as only accepting strings :slightly_smiling_face:

jgroth: We can also just add to the documentation that the values need to be serialized or they won't be saved :smile:

civing: Should we go with Johans proposal? Leave it as is but document that data needs to be serializable?

adrianschmidt: Yeah, let's go with `set<T = any>(key: string, data?: T): Promise<void>;` then 👍

src/userpreferences/repository.ts
adrianschmidt: Leave this as it is, since we keep the generic type 👍

src/userpreferences/types.ts
```diff
@@ -0,0 +1,21 @@
+import { PlatformServiceName } from '../core/platform';
+import { UserPreferencesRepository } from './repository';
+
+const SERVICE_NAME = 'state.user-preferences';
```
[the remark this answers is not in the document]
adrianschmidt: Well just go with `userPreferences` if that's ok?

src/userpreferences/repository.ts
```diff
@@ -15,13 +15,14 @@ export interface UserPreferencesRepository {
      * Set user data
      *
      * @param key - name of the key for the data
-     * @param data - the data to save, will delete the data if undefined
+     * @param data - the data to save, will delete the data if undefined. However
+     * @note the data need to be serializable
```
adrianschmidt: I think you might have discovered this yourself, since you changed it in another fixup, but an `@note` can't go inside the description for an `@param` like this, because it becomes a tag of its own and ends up in a separate part of the documentation section:

image

src/userpreferences/repository.ts
```diff
@@ -0,0 +1,27 @@
...
+     * @param key - name of the key for the data
+     * @param data - the data to save, will delete the data if undefined
+     */
+    set(key: string, data?: string): Promise<void>;
+
+    /**
+     * Gets the schema for the user preferences.
+     *
+     */
+    getSchema(): object;
```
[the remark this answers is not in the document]
adrianschmidt: ⚡

PR #592, opened 2023-01-19
feat(Task): add options to hide notifications from task

terryilebode6:
the aim of this PR is to add the arguments to the task repository interface that allow us to disable toasts and banners from tasks


src/task/repository.ts
```diff
@@ -13,13 +13,18 @@ export interface TaskRepository extends StateRepository {
...
      */
     create(
         url: string,
         data: any,
         cancelAction?: boolean,
-        message?: string
+        message?: string,
+        hideBanner? : boolean,
+        hideToast? : boolean
+
```
jgroth, line 27: I think there are too many arguments to this method now. Four were already stretching it so adding another two makes it very weird to use. For someone reading the code, it will be very hard to understand what all these booleans are doing :smile: 
```ts
service.create('/foo', 'data', true, 'hello', false, true);
```

I think a better approach is to overload the `create` method with a better signature, and maybe we can even deprecate the one we have today. Perhaps something like this?

```ts
create(url: string, data: unknown, options: TaskOptions);
```

jgroth (commented):
Nice work, I think we need to improve the API though :slightly_smiling_face: 

There are also some formatting problems, and since this updates the API, you need to run the `npm run api:update` command to get the build to pass :slightly_smiling_face:

PR #597, opened 2023-01-25
docs(package): added guidelines for release statuses

jgroth:
Added some guidelines on how consumers should handle the different release statuses, it will be visible at the top of the front page of the documentation. This is what made most sense to me, any feedback is welcome :smile:


jgroth:
After writing this I found the official meaning of the tags according to TSDoc, but it's pretty similar to what I wrote :slightly_smiling_face: 
https://tsdoc.org/pages/tags/beta/

src/index.ts
```diff
@@ -2,6 +2,21 @@
...
+ * following statuses
+ *
+ * - `@public` - released and ready to be used. The interface is stable and
+ * should not have any breaking changes without triggering a new major release.
+ * - `@beta` - the interface is quite stable and can be used by those willing
+ * to accept breaking changes. Breaking changes should be kept to a minimum and
+ * will not trigger a new major release. Feedback on the interfaces are welcome.
+ * - `@alpha` - still under development and is likely to have frequent breaking
+ * changes without notice.
+ * - `@deprecated` - Can be used for the time being, but will be removed in the
```
Befkadu1, line 15:
```suggestion
 * - `@deprecated` - can be used for the time being, but will be removed in the
```

Befkadu1:
Awesome 👏

src/index.ts
adrianschmidt: I think it's weird to start the first sentence with a lowercase letter and then end it with a period.

(It's even worse when single sentence descriptions are started with a lowercase letter and just ends without a period, and then some descriptions need more than one sentence and therefore need periods and uppercase letters 🤢)

I think the only way to be consistent, is to start sentences with uppercase letters, and end them with some kind of punctuation.

Befkadu1: Yep,  I am not good in written English but I checked the consistency 😆

adrianschmidt: I can fix, so we can get this one merged 👍

src/index.ts
```diff
@@ -5,12 +5,13 @@
...
+ * - `@public` - Released and ready to be used. The interface is stable and
  * should not have any breaking changes without triggering a new major release.
- * - `@beta` - the interface is quite stable and can be used by those willing
- * to accept breaking changes. Breaking changes should be kept to a minimum and
- * will not trigger a new major release. Feedback on the interfaces are welcome.
- * - `@alpha` - still under development and is likely to have frequent breaking
+ * - `@beta` - The interface is fairly stable and can be used by those willing
+ * to accept breaking changes without notive. Breaking changes may still happen
+ * without notice, and without triggering a new major release. Feedback on the
+ * interfaces are welcome.
```
adrianschmidt, line 13: I made `@beta` sound a bit less attractive to use.

Befkadu1 (dismissed):

src/index.ts
adrianschmidt: ⚡

src/index.ts
```diff
@@ -2,6 +2,22 @@
  * This is the public API for Lime Web Components. It contains helper functions
  * and interfaces for all services available in the Lime CRM platform.
  *
+ * Everything that is exported from this package is marked with one of the
+ * following statuses
+ *
+ * - `@public` - Released and ready to be used. The interface is stable and
+ * should not have any breaking changes without triggering a new major release.
+ * - `@beta` - The interface is fairly stable and can be used by those willing
+ * to accept breaking changes without notive. Breaking changes may still happen
```
Befkadu1, line 11:
```suggestion
 * to accept breaking changes without notice. Breaking changes may still happen
```

adrianschmidt: Ah! Nice! I'll fix!

adrianschmidt: Fixed and amended to the existing commit.

Befkadu1 (approved):

adrianschmidt (approved):

PR #598, opened 2023-01-25
Use async when saving

Befkadu1:
2 changes has been done.
1. Changing `userdata` from `void` -> `Promise<void>` which is the public service 
2. Changing `userpreferences` from `void` -> `Promise<void>` which is in alpha 

However the changes in `userdata` might not be a breaking change. It won’t break anything. What do you think @adrianschmidt ?


src/userdata/repository.ts
```diff
@@ -28,6 +28,7 @@ export interface UserDataRepository extends StateRepository {
      *
      * @param key - name of the key for the data
      * @param data - the data to save, will delete the data if undefined
+     * @returns a promise that will be resolved when the data is saved
      */
-    set<T = any>(key: string, data?: T): void;
+    set<T = any>(key: string, data?: T): Promise<void>;
```
adrianschmidt, line 33: > 2 changes has been done.
> 
> 1. Changing `userdata` from `void` -> `Promise<void>` which is the public service
> 2. Changing `userpreferences` from `void` -> `Promise<void>` which is in alpha
> 
> However the changes in `userdata` might not be a breaking change. It won’t break anything. What do you think adrianschmidt ?

I don't think this change ought to be breaking anything for a consumer of the service, but I assume it _will_ be breaking for the implementer of the service… 🤔

Input @jgroth?

adrianschmidt: Workorder uses this service, so I can test it against that 👍

adrianschmidt: Ah, workorder is using a really old version of lime-web-components, so it's still importing stuff from `@limetech/lime-web-components-interfaces`…

But at least that means they already have breaking changes to deal with when updating.

I think we should still mark this as breaking, since it will be for implementations of the service, but we can write that it's shouldn't affect consumers. I can take a stab at writing something, and see if I can make it make sense 😅

jgroth: Yeah any existing consumer would never rely on a return value, so if it starts returning one they would just ignore it.

It will be breaking for the implementer though, but you can change it to `void | Promise<void>` to get around that, but it's a bit ambiguous. Another alternative is to add an optional overload that returns the promise version, and deprecate the one with `void`.

You could however just change the return type of the new service without changing the one on the old one we already have. The implementation just have to rely on the implementation of the other service (which returns a promise), rather than the interface.

jgroth: > I think we should still mark this as breaking

I would rather not do a breaking change for a change like this already since we made one not too long ago

adrianschmidt: What's the problem with making a breaking change if we communicate that it's not breaking for consumers?

Befkadu1: > What's the problem with making a breaking change if we communicate that it's not breaking for consumers?

Is there any label like "Note" instead of using "BREAKING CHANGE"?

Befkadu1: Then let us go with the optional overload and deprecating the old version.

adrianschmidt: > Is there any label like "Note" instead of using "BREAKING CHANGE"?

No, unfortunately there isn't. I would like to add something like that, but haven't had the time. It would require adding a feature to the changelog plugin for semantic-release.

(But, if a change is breaking, we should _always_ mark it as breaking.)

adrianschmidt: An optional overload won't help it seems, because for some reason, if one overload is optional, typescript requires all overloads to be optional, so we'd have to change the existing function to be optional, which would be a breaking change.

adrianschmidt: I don't really see the value in making the code more complex just to avoid a breaking change, if the breaking change doesn't actually require anyone currently using the service to change anything.

I really don't understand how this change is more problematic than the _massively_ breaking change we made in September? 🤷‍♂️

etc/lime-web-components.api.md
```diff
@@ -783,7 +783,7 @@ export interface UserCoworker {
 export interface UserDataRepository extends StateRepository {
     get<T = any>(key: string): T;
```
adrianschmidt, line 784: While we're at it, shouldn't the getters also return a promise? 🤔

src/userdata/repository.ts
Befkadu1: Yea, I have mentioned that [here](https://github.com/Lundalogik/limeclient.js/pull/303#discussion_r1085524448). So the implementer is using `async` while we don't have async in the interface 😆 . Amazing! I don't think we have implementer other than limeclient.js. Then we can remove the breaking change tag

adrianschmidt: > Then we can remove the breaking change tag

Already done 😁

etc/lime-web-components.api.md
Befkadu1: I don't think it returns a promise. We fetch the userdata when it loads [here](https://github.com/Lundalogik/limeclient.js/blob/5db34e6970d03c7bc37c0808be03c4484813fd57/src/userdata/repository.ts#L48) so no need of promise for get request here

adrianschmidt: Ah 👍

src/userdata/repository.ts
jgroth: > What's the problem with making a breaking change if we communicate that it's not breaking for consumers?

Breaking changes in general are annoying because consumers have to check what has changed and make a decision if they can/want to upgrade or not. It also makes it harder for them to upgrade if we release a fix for something important, because there might be a breaking change merged in between that will force them to upgrade just to get the important fix. If possible, I think it's better to introduce non breaking changes and then collect breaking changes and release them all at once like we have done before. 

In this case, there is no indication from either the interfaces themselves or the documentation that these two services are depending on each other, so it is possible to change just one of them without affecting the other. If the implementation uses the other service internally (like we are), so be it. But I think you already discovered that :slightly_smiling_face:

src/userdata/repository.ts
```diff
@@ -28,6 +28,7 @@ export interface UserDataRepository extends StateRepository {
      *
      * @param key - name of the key for the data
      * @param data - the data to save, will delete the data if undefined
+     * @returns a promise that will be resolved when the data has been saved
```
adrianschmidt, line 31: I changed this from "will be resolved when the data is saved" to "will be resolved when the data has been saved", to make the intention as clear as possible.

src/userpreferences/repository.ts
```diff
@@ -18,8 +18,9 @@ export interface UserPreferencesRepository {
      *
      * @param key - The key to store the value under.
      * @param data - The data to save. Leave undefined to instead delete the currently stored value.
+     * @returns a promise that will be resolved when the data has been saved
```
adrianschmidt, line 21: Same change from "is" to "has been" here.

adrianschmidt (approved):

PR #605, opened 2023-02-01
feat(dateformatter): add new interface `DateFormatter`

adrianschmidt:
fix: https://github.com/Lundalogik/crm-feature/issues/3180


src/dateformatter/types.ts
```diff
@@ -0,0 +1,34 @@
...
+ * @group Localization
+ */
+export type DateType =
+    | 'datetime'
+    | 'date'
+    | 'time'
+    | 'week'
+    | 'month'
+    | 'quarter'
+    | 'year';
```
jgroth, line 34: I think we can move these to `src/limetype/property.ts`. We can refactor the `PropertyType` in there and break it up to into something like this instead of declaring them again
```ts
export type PropertyType =
    | 'string'
    | 'text'
    | 'phone'
    | DateType
```

src/dateformatter/dateformatter.ts
```diff
@@ -0,0 +1,20 @@
...
+export interface DateFormatter {
+    /**
+     * Format the given date as the given type, in the locale set in the
+     * user preferences.
+     *
+     * @param date - The date to format.
+     * @param type - The type to format the date as.
+     * @returns A string representation of the date.
+     */
+    get(date: Date, type: DateType): string;
```
jgroth, line 19: All date properties on `LimeObject` will be strings, so in order to use this with those properties the user would have to first convert it into a `Date` object, which can make it harder to use than necessary. I think we could still allow `Date` here to make it usable in other scenarios, but I think we should accept ISO formatted date strings as well

src/dateformatter/dateformatter.ts
```diff
@@ -0,0 +1,20 @@
+import { DateType } from './types';
+
+/**
+ * Service for formatting dates according to the locale set in the
+ * user preferences.
+ *
+ * @alpha
+ * @group Localization
+ */
+export interface DateFormatter {
```
jgroth, line 10: How about naming this `DateTimeFormatter` since it handles time as well? :slightly_smiling_face:

src/dateformatter/dateformatter.ts
jgroth: Also, how do you feel about the naming here and `get` vs `format`?

src/dateformatter/dateformatter.ts
```diff
@@ -0,0 +1,20 @@
...
+    /**
+     * Format the given date as the given type, in the locale set in the
+     * user preferences.
```
jgroth, line 13: Perhaps we should not couple this too much to the user preferences :thinking: 
We could add an optional `locale` argument to the method and fall back to the user preferences if it is not provided :thinking:

src/dateformatter/dateformatter.ts
jgroth: If we add another argument, perhaps we should do it like this in case we want other options in the future
```ts
get(date: string | Date, options: FormattingOptions): string;
```

src/dateformatter/types.ts
adrianschmidt: Yes! Nice input! I've seen `PropertyType` in lime-crm-components, and reflected on the fact that this duplication was annoying, but I didn't realise then that `PropertyType` actually comes from lime-web-components. I'll fix! 👍

src/dateformatter/dateformatter.ts
adrianschmidt: You know you can start multiple threads on the same line, right? 😄

I'll make three threads of these comments 👍

adrianschmidt, line 19: jgroth wrote:
> All date properties on `LimeObject` will be strings, so in order to use this with those properties the user would have to first convert it into a `Date` object, which can make it harder to use than necessary. I think we could still allow `Date` here to make it usable in other scenarios, but I think we should accept ISO formatted date strings as well

adrianschmidt, line 19: jgroth wrote:
> If we add another argument, perhaps we should do it like this in case we want other options in the future
> 
> ```ts
> get(date: string | Date, options: FormattingOptions): string;
> ```

adrianschmidt: Good point. I'll update it 👍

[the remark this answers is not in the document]
adrianschmidt: Yeah, `format` is probably better. I'll change it 👍

adrianschmidt: Hmm…

I see the point, and I agree to an extent. At the same time, `type` is required, so we'd be requiring people to supply an object every time. That seems a bit annoying.

Allowing the user to pass in a locale, as you mention in another comment, seems like a good idea. I would suggest one of these, probably the latter:

```ts
export interface FormattingOptions {
    locale?: string;
}

    get(date: Date, type: DateType, options?: FormattingOptions): string;
```

```ts
export interface FormattingOptions {
    type: DateType;
    locale?: string;
}

    get(date: Date, options: DateType | FormattingOptions): string;
```

Thoughts?

src/dateformatter/dateformatter.ts
adrianschmidt: I would have argued that a `Date` in javascript can only include _both_ a date and a time, and thus `DateFormatter` is enough… but since `Intl.DateTimeFormat` is named as it is, I guess that sets a different precedent. I'll update it 👍

src/dateformatter/dateformatter.ts
jgroth: I like the second one better I think. By having `DateType | FormattingOptions` we get around the annoyance you mentioned where it would have to be used as `get(date, { type: 'month' })` instead of `get(date, 'month')`

src/dateformatter/dateformatter.ts
adrianschmidt: Good point. This touches upon the interface signature, so I replied here: https://github.com/Lundalogik/lime-web-components/pull/605/files#r1094316796

I'll resolve this, so we'll have a unified discussion in the other thread.

src/dateformatter/types.ts
adrianschmidt: Hmm… `PropertyType` uses `timeofday` instead of `datetime`. We obviously can't change `PropertyType`. We _could_ decide to use `timeofday` instead of `datetime` in the `DateTimeFormatter` interface but that would mean we couldn't pass this function straight into `limel-date-picker` (via the new optional property for supplying a date-formatter function, which we are planning to add), because `limel-date-picker` uses `datetime`. In general, I think `datetime` and `timestamp` are widely used names for the concept "a date and a time" (where `timestamp` probably tends to suggest that at least seconds, and probably milliseconds, are considered important), but I don't think `timeofday` is really a thing outside Lime 😅

Maybe the easiest thing is to allow both `datetime` and `timeofday`, and simply state in the docs that implementations of the service are expected to handle them the same?

src/dateformatter/dateformatter.ts
adrianschmidt: ⚡

src/dateformatter/dateformatter.ts
adrianschmidt: ⚡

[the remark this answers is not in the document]
adrianschmidt: ⚡

adrianschmidt: ⚡

src/dateformatter/types.ts
jgroth: `timeofday` is only time I think, `datetime` is for both time and date. I think the confusing thing is that we also have `time` :thinking: :smile:

adrianschmidt: `PropertyType` does not include `datetime`: https://github.com/Lundalogik/lime-web-components/blob/bd11472279f26544c427a37b56eaee6a322b2b30/src/limetype/property.ts#L42

jgroth: Aha! Yeah maybe that's what `time` was used for, and then we needed to show only time and came up with `timeofday` just to be backwards compatible :upside_down_face:

adrianschmidt: Pushed a fixup for this now. At the time of writing it's the latest one. Have a look and see if you like it or not…

I guess we'll want to accept `timeofday` regardless, as people will probably be getting that from their properties, and then will want to just pipe that into the formatter service…?

adrianschmidt: Although I guess we already do expose it through `PropertyType` 😭

jgroth: Yeah we need to accept all types we can get from the backend, and the backend will never return `datetime`. If our other components (like Lime Elements) use `timestamp` or `datetime` I think we need to handle that conversion internally in our components/services

adrianschmidt: Ok, so at least that means we can't use some union of the relevant `PropertyType`s and the types used by `limel-date-picker`, because both have the value `time` but it means completely different things in each.

So we need to decide.

While I'm typically biased towards the decisions previously made in lime-elements (I wonder why that might be? 😝), I'm actually willing to lean towards Lime's `PropertyTypes` here… although, they still don't have a value for `week`, right? That's why we always need to have that extra `format` param alongside the `type`… I must admit I'm _extremely_ reluctant to propagating that awfulness any further…

adrianschmidt: > Yeah we need to accept all types we can get from the backend, and the backend will never return `datetime`. If our other components (like Lime Elements) use `timestamp` or `datetime` I think we need to handle that conversion internally in our components/services

Yup, I'm on board with this. We just need to solve the `week` format. But I guess we can basically do what I did in the last fixup, but instead of adding `datetime`, we simply add `week`, and then let the implementation of the service handle that…

adrianschmidt: And, we _definitely_ need to document `DateTimePropertyType` and explain what the heck the different values mean, since that's anything but obvious… 😖

src/datetimeformatter/types.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * The possible types a date can be formatted as.
+ *
+ * `datetime` is an alternative name for `timeofday`. Implementations of the
+ * `DateTimeFormatter` service are expected to treat these two values the same.
+ *
+ * @alpha
+ * @group Localization
+ */
+export type DateTimeType = 'datetime' | DateTimePropertyType;
```
jgroth, line 33: Do we need to include `datetime`? It's not something that exist today on `LimeProperty` so if we include it, it's to make the service more userfriendly when used on it's own. If we include it, I think we should change the documentation and say it's the same as the `time` property

jgroth (commented):
Great work! I think it looks good, just not sure about the `datetime` type :smile:

src/dateformatter/types.ts
jgroth: Ah, yeah `week` is missing from the backend as well so that will also never be returned. Noticed that just now... :smile:
How is that handled today? :thinking: Is it configured in lime-admin?

adrianschmidt: Yeah, I think you still have to set `format="week"` in the properties sections in lime-admin… That's how it used to work at least.

I still cannot fathom why this specific date format is apparently treated differently from all others in LISA. (I assume it's different there, since we don't just get the `type` as with everything else…)

jgroth: In LISA there is also a date option called "Four months", so a third of a year :smile:

src/datetimeformatter/types.ts
adrianschmidt: Removed again. Let's keep this as part of the long discussion though 🙂

src/dateformatter/types.ts
adrianschmidt: Just pushed another fixup, this time removing `datetime` and instead adding `week`.

I guess if people use the `PropertyType` value to send in here, they need to look at if the value is `type`, and the config for the field says `format="week"`, they need to send in `week` instead of `time`…

(It would be really nice to have `week` _fixed_ in the core…)

jgroth: Yeah, it will be a bit harder right now for users to use the `week` type since they have to configure it manually instead of just using `property.type`, but I think that is the best we can do right now :slightly_smiling_face:

jgroth: > Yeah, also not covered here. I wonder if we'd actually get that as a type though, or if it would be just like week

It will be just like `week`. All the types we get from the backend are the ones listed in `PropertyType`

jgroth (approved):

PR #607, opened 2023-02-02
docs(DateTimePropertyType): add some documentation about the less intuitive of the date types

adrianschmidt:
Hopefully my description is correct… please double-check! 🙏 😅


src/limetype/property.ts
```diff
@@ -36,6 +36,14 @@ export interface Option {
...
 
 /**
+ * The subset of `PropertyType`s that represent a date or time of some kind.
+ *
+ * **NOTE**<br>
+ * For legacy reasons:
+ * - the type `time` represents a value with date _and_ time
+ * - the type `timeofday` represents a value with _only_ time
+ * - there is no type for "week"; instead week properties use the type `time` and are configured as "week" in the applicable view config
+ *
```
jgroth, line 46: I'm even more confused now. I thought this was true (and it probably is) but I found some really weird stuff when I was digging through some code for confirmation.

In the translation files in the webclient I found this which indicates that `time` is just time.

```
msgid "lime-datetime.format.datetime"
msgstr "L - LT"

msgid "lime-datetime.format.date"
msgstr "L"

msgid "lime-datetime.format.time"
msgstr "LT"
```

This does not mean that the property with `type="time"` is used with this formatting string however, we could have some mapping in the code to this translation key.

I also found this in some very old angular code which seems to use these, and also has some internal mapping from `timeOnly` (which I _think_ was something we only used in the angular code). I did not find any mapping for `time` in here
```ts
switch (format) {
    case 'datetime':
        momentFormat = $translate.instant(
            'webclient.lime-datetime.format.datetime'
        );
        break;
    case 'date':
        momentFormat = $translate.instant(
            'webclient.lime-datetime.format.date'
        );
        break;
    case 'timeOnly':
        momentFormat = $translate.instant(
            'webclient.lime-datetime.format.time'
        );
        break;
}
```

This was found this in some "newer" angular code, which suggests that `time` in fact **is** both date and time.
```ts
const dateTypeMap = {
    time: 'datetime',
    timeOnly: 'time',
};
```

This is also supported by this snippet from crm-components, so this is code that is running today.
```ts
if (this.type === 'time' && !format) {
    format = 'datetime';
}
```

The `timeofday` property from the backend was added quite recently, and I know we did it for a reason. There was back then no way to map fields with only time from LISA in the Python backend and we needed a way to do that.

jgroth: So... TLDR
Yes, I think this is a correct description after all the confusion :smile:

adrianschmidt: Omg… 😩

I guess the simplest way to know for sure is to do some `console.log`s in appropriate places. I'll be tinkering with the formatter today, and I have all kinds of date and time fields set up on `company` in my dev-db, so I should be able to figure out the current state of affairs, without having to know exactly how we got in this mess 😅

Thanks for the help!

jgroth: I found this in limeclient as well. This is probably what is most close to the truth :smile: This was the only place I found any reference whatsoever to the `timeofday` type
```ts
function getDateType(type: string) {
    const mapping: any = {
        time: 'datetime',
        date: 'date',
        year: 'year',
        quarter: 'quarter',
        month: 'month',
        timeofday: 'time',
    };

    return mapping[type];
}
```

src/limetype/property.ts
```diff
@@ -36,6 +36,14 @@ export interface Option {
 }
 
 /**
+ * The subset of `PropertyType`s that represent a date or time of some kind.
+ *
+ * **NOTE**<br>
+ * For legacy reasons:
+ * - the type `time` represents a value with date _and_ time
+ * - the type `timeofday` represents a value with _only_ time
+ * - there is no type for "week"; instead week properties use the type `time` and are configured as "week" in the applicable view config
```
adrianschmidt, line 45:
```suggestion
 * - there is no type for "week"; instead week properties use the type `date` and are configured as "week" in the applicable view config
```

adrianschmidt: The `type` is `date`, but I still need to find out what to write about the `format`…

adrianschmidt: I guess this is good enough. It's much better than before…

